Issue #34, opened 2015-04-13
The client objects don't appear to be threadsafe.


harlowja:
It doesn't appear that the usage of the same socket (and/or client) by different threads is safe. It would be great to either mention that in the docs or make it thread safe.

A possible variation that I think makes this safe is @ https://review.openstack.org/#/c/173101/ (feel free to take that code or adjust it... as u want).

harlowja:
Or a connect/socket pool (that is thread-safe) would be even better... that each call into a a command grabs/releases a socket from that pool (and therefore they can't overlap no matter what).

cgordon:
I've commented in the PR on this, and agree that it warrants documentation, which I'll add once we've resolved the PR.

PR #35, opened 2015-04-14
Make each client thread safe

harlowja:
- Add a object pool that can create sockets in a thread
  safe manner (and release and destroy them as/when needed).
- Avoid using a client 'buffer' and use a per-function/method
  buffer instead to avoid having multiple threads writing into
  the same buffer.

Fixes issue #34

### DIFF
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -695,12 +695,12 @@ class Client(object):
             try:
                 buf = b''
                 sock.sendall(cmd)
-    
+
                 result = {}
                 while True:
                     buf, line = _readline(sock, buf)
                     self._raise_errors(line, name)
-    
+
                     if line == b'END':
                         return result
                     elif line.startswith(b'VALUE'):
@@ -715,10 +715,10 @@ class Client(object):
     
                         buf, value = _readvalue(sock, buf, int(size))
                         key = checked_keys[key]
-    
+
                         if self.deserializer:
                             value = self.deserializer(key, value, int(flags))
-    
+
                         if expect_cas:
                             result[key] = (value, cas)
                         else:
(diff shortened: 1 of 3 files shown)


cgordon:
Hey Josh, thanks for the pull request! I've been meaning to add connection pooling (and consistent hashing) to pymemcache, but time is always in short supply (and we use an external proxy for that at Pinterest, so we haven't needed it).

One of the design goals of the Client object was that it be easy to compose so we could add things like pooling and hashing later, without having to change it. For instance, my preference for pooling would be to create a separate PooledClient that maintains a pool of Client objects and implements each method like this:

```
def add(self, key, value, expire=0, noreply=True):
    with self.client_pool.get_and_release() as client:
        return client.add(key, value, expire=expire, noreply=noreply)
```

This means that users of Client don't have to pay any extra overhead for pooling (however small), but anyone who wants can use pooling with the same interface (it's a drop-in replacement) with only a small extra overhead (one extra method call). In the future, if we also add a HashingClient (or something similar) we could have PooledClient keep a pool of those, and have the HashingClients delegate to Clients, keeping everything nicely orthogonal. What do you think? Is there something about this particular change that precludes that approach?

Separately, converting buf to a local variable looks both safe and more straightforward than having it as an instance variable. I cannot remember, now, why I wrote it as an instance variable originally, but it's probably vestigial. If you want to make that change, it would be great, or I can do it after this PR.

CG.

harlowja:
That seems fair, I should be able to reuse this little object pool thingy for that.

harlowja:
Will attack that way of doing it tommorow :-P

harlowja:
Reworking, will open new PR.

PR #36, opened 2015-04-14
Remove instance local buffer

harlowja:
It doesn't seem needed to have instance local-side effects
especially around a shared buffer (having it exists makes
thread-safety and/or thread-safe clients that much harder to
achieve).


cgordon:
Looks good, thanks for the fix!

cgordon:
I'll wait to tag this and upload it to pypi for a couple days so we can get your other change in as well (no hurry!).

PR #37, opened 2015-04-14
Create a pooled client class

harlowja:
- Add a object pool that can create clients in a thread
  safe manner (and release and destroy them as/when needed).


harlowja:
Initial attempt at this, can add docs/tests if needed.

harlowja:
Tests added, no docs, but merge as u wish :-P

cgordon:
This looks great! Thanks for cleaning up the unit tests.

PR #38, opened 2015-04-15
Add harlowja to credits in README

harlowja:
Might as well add myself :smile:


cgordon:
Ha :)

I was going to take care of this when I bumped the version and uploaded to pypi later this week, but this makes it easier!

PR #41, opened 2015-04-16
Allow the pool lock to be specialized

harlowja:
It can be quite useful to allow the lock that the pool
uses to be specified to be something other than the default
threading lock class/type. This can be especially useful when
eventlet (or gevent or other) are used which have there own
special lock type that connects into there event-loop.

### DIFF
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -826,13 +826,14 @@ class Client(object):
 class PooledClient(object):
     """A thread-safe pool of clients (with the same client api).
 
-    :param max_pool_size: maximum pool size to use (going about this
-                          amount triggers a runtime error), by default this
-                          is 2147483648L when not provided.
-    :param lock_generator: a callback/type that takes no arguments that will
-                           be called to create a lock or sempahore that can
-                           protect the pool from concurrent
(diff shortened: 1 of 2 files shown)


pymemcache/client.py
```diff
@@ -836,7 +836,10 @@ def __init__(self,
                  ignore_exc=False,
                  socket_module=socket,
                  key_prefix=b'',
-                 max_pool_size=None):
+                 max_pool_size=None,
+                 # Allow this to be a different style of lock (eventlet
+                 # lock/semaphore for example, that works with greenthreads).
```
jparise: This type of documentation generally goes in a docblock.

harlowja: Ya. I'll add those into a docstring/block on the class.

cgordon:
This is pretty nit-picky, but we've been using the Google style guide for documentation, and I have a mild preference that we stay consistent with that:

https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Comments#Comments

cgordon:
Apologies! I lost track of this change, merging now..

PR #42, opened 2015-04-16
Add helper that ensures secondary exceptions don't get propagated

harlowja:


cgordon:
I like this context manager, and am going to steal it for other things, but it doesn't work very well here. The problem is that the close() method never throws exceptions (it swallows them), so this will never be triggered. One possibility is that you wrap the body of the close() method with this context manager, after removing the initial assertion (since we can call close with no active exceptions).

harlowja:
Fair with me, good point, maybe not really needed :-P

harlowja:
https://github.com/openstack/oslo.utils/blob/master/oslo_utils/excutils.py#L30 is actually a little bit better btw...

PR #44, opened 2015-05-04
add shardingclient

duanhongyi:
Base on consistent hash, here's an example::

    from pymemcache.client import PooledClient, ShardingClient
    c1 = PooledClient(("10.1.15.12", 11211))
    c2 = PooledClient(("10.1.15.13", 11211))
    client = ShardingClient([c1, c2])
    client.set('key', {'a':'b', 'c':'d'})
    result = client.get('key')

### DIFF
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -1219,19 +1219,19 @@ class ShardingClient(object):
 
     def stats(self, *args):
         results = {}
-        for client in self.clients:
+        for client in self.servers.values():
             results.update(client.stats(*args))
         return results
 
     def flush_all(self, delay=0, noreply=True):
         ok = True
-        for client in self.clients:
+        for client in self.servers.values():
             if not client.flush_all(delay, noreply):
                 ok = False
         return ok
 
     def quit(self):
-        for client in self.clients:
+        for client in self.servers.values():
             client.quit()
 
     def __setitem__(self, key, value):
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -1047,7 +1047,8 @@ class ShardingClient(object):
     def get_client(self, key):
         if self.black_list:
             self.sync_black_list()
-        return self.servers[self.ring.get_node(key)]
+        client = self.servers[self.ring.get_node(key)]
+        return client
 
     def close(self):
         for server in self.servers.values():
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -69,10 +69,11 @@ Best Practices:
 
 __author__ = "Charles Gordon"
 
-
+import time
 import socket
 import six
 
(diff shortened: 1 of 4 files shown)


cgordon:
I'm on vacation right now, but will be back this coming weekend, and will review this then. Thank you for the commit!

duanhongyi:
@cgordon

I will fix test case for py3.3

pymemcache/client.py
```diff
@@ -113,11 +114,13 @@ def json_deserializer(key, value, flags):
 
 
 class MemcacheError(Exception):
+
     "Base exception class"
     pass
 
 
 class MemcacheClientError(MemcacheError):
+
```
cgordon: Please remove these extra empty lines in the Error classes (see the five below).

pymemcache/client.py
```diff
@@ -999,6 +1009,256 @@ def __delitem__(self, key):
         self.delete(key, noreply=True)
 
 
+class ShardingClient(object):
+
+    def __init__(self, clients, black_list_timeout=30):
```
cgordon: Please document this method similarly to the constructors for PooledClient and Client.

pymemcache/client.py
```diff
@@ -999,6 +1009,256 @@ def __delitem__(self, key):
         self.delete(key, noreply=True)
 
 
+class ShardingClient(object):
```
cgordon: The client.py file is getting rather large. Let's move ShardingClient to a new file called sharding_client.py. I'll do the same for PooledClient in a separate commit.

pymemcache/ring.py
```diff
@@ -0,0 +1,173 @@
...
+from bisect import bisect
+
+if sys.version_info >= (2, 5):
+    import hashlib
+    md5_constructor = hashlib.md5
+else:
+    import md5
+    md5_constructor = md5.new
+
+class HashRing(object):
```
cgordon: Did you write this code? If not, where did you get it? I just need to double check the licensing.

duanhongyi: @cgordon 

Hash_ring source code from the end of 2012 no longer updated, it does not support py3k, so I update hash_ring.

duanhongyi: @cgordon

From here:
https://pypi.python.org/packages/source/h/hash_ring/hash_ring-1.3.1.tar.gz

duanhongyi:
@cgordon

merge && modified

sontek:
@duanhongyi I started implementing an alternative to this in #59, want to take a look?  I also documented some of the reasoning behind it in #52

duanhongyi:
close && see:

https://github.com/pinterest/pymemcache/pull/59

PR #45, opened 2015-05-07
Retry on EINTR while sock.recv() and sock.sendall()

methane:
See https://www.python.org/dev/peps/pep-0475/

### DIFF
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -70,6 +70,7 @@ Best Practices:
 __author__ = "Charles Gordon"
 
 
+import errno
 import socket
 import six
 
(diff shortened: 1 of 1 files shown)


cgordon:
Is there any way to test these changes? I think I understand the rationale, but I haven't had time to dig through the relevant Python C code for socket handling to double check that it isn't already handling retries on EINTR.

methane:
@cgordon What type of test do you want?
It's easy to add `OSError(errno.EINTR)` to `MockSocket`.  But it doesn't check about underlaying socket doesn't retry.
May I make sample script for demonstrate EINTR separated from unittest?

methane:
https://gist.github.com/methane/59781ec74e00402e4c6b

In this script, client.recv() raises EINTR, but sendall() doesn't.

methane:
I've made #47 since I don't like branch name "patch-2".

cgordon:
Should I close this PR in favor of #47 ?

PR #47, opened 2015-05-11
Retry when sock.recv() raises EINTR

methane:
On Python 2.7 and 3.4, `sock.recv()` returns EINTR when it interrupted by signal.
https://gist.github.com/methane/59781ec74e00402e4c6b

Some web server uses signal to manage workers. So pymemcache should retry recv() when EINTR.

(Python 3.5 retries automatically. See https://www.python.org/dev/peps/pep-0475/ )

### DIFF
--- a/pymemcache/client.py
+++ b/pymemcache/client.py
@@ -69,7 +69,7 @@ Best Practices:
 
 __author__ = "Charles Gordon"
 
-
+import errno
 import socket
 import six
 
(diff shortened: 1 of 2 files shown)


methane:
I'm surprised by test for 767c47e was succeeded.
It fails on my Mac.

cgordon:
Is the test working correctly on your Mac now? If so, I can merge this.

methane:
> Is the test working correctly on your Mac now? If so, I can merge this.

Yes.

cgordon:
Thanks!

PR #50, opened 2015-06-20
Add docs

sontek:
This sets up sphinx for documentation, you can see what this looks like here:

http://pymemcache.readthedocs.org/en/latest/apidoc/pymemcache.html#module-pymemcache.client
http://pymemcache.readthedocs.org/en/latest/getting_started.html#basic-usage


cgordon:
This is awesome! Thanks!

PR #51, opened 2015-06-20
Switch to pytest

sontek:
This switches you to use pytest instead of nose which is a much more active project and has a much better reporting mechanism.  This also enables displaying code coverage by default when running tests.

#55 should be merged first.


cgordon:
Can you give me more context for this change? I'm not familiar with pytest, and have used nose at both jobs for which I used Python. Can you point me at some comparisons somewhere, or just give me a quick synopsis of why pytest is better?

sontek:
@cgordon Yeah, sorry I should have marked this one as a work in progress.  You will see the benefit once I have the integration tests ported over as well. I'll show some before/after output as well.

sontek:
@cgordon ok, check out the integration tests now

sontek:
So the primary reasons you would want to switch:

1. More pythonic, you can use `assert foo == bar` instead of `tools.assert_equal(foo, bar)`
2. py.test is more actively maintained and by an extremely strong maintainer (guy who also created tox and devpi)
3. Better error report, for example look at this:

```python
def test_dict_difference():
    assert {'foo': 'bar'} == {'bar': 'foo'}
```

In nosetests:
```
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/sontek/venvs/pymemcache/src/test_foo.py", line 2, in test_dict_difference
    assert {'foo': 'bar'} == {'bar': 'foo'}
AssertionError
```

in pytest:
```
    def test_dict_difference():
>       assert {'foo': 'bar'} == {'bar': 'foo'}
E       assert {'foo': 'bar'} == {'bar': 'foo'}
E         Left contains more items:
E         {'foo': 'bar'}
E         Right contains more items:
E         {'bar': 'foo'}
E         Full diff:
E         - {'foo': 'bar'}
E         + {'bar': 'foo'}
```
4. For setting up fixtures (dependencies of your tests) you have the ability to use dependency injection to send many different parameters to your test functions (check the integration test file to see this in action)

cgordon:
Any reason to keep pymemcache/test/integration.py around? Other than that, I'm ready to merge here. Pytest looks nicer, and I really like the output on failed tests.

sontek:
@cgordon oh, nope.  Just missed the delete on my commit.  I'll fix

Issue #52, opened 2015-06-20
Add support for a pluggable consistent hashing system


sontek:
Options:

1. https://pypi.python.org/pypi/hash_ring
2. https://github.com/RJ/ketama/tree/master/python_ketama
3. https://github.com/ewdurbin/clandestined-python

Ketama is the one that `pylibmc` supports but since the protocols between the clients aren't 100% compatible there isn't any reason to completely follow their lead.

sontek:
After doing a bunch of testing I've decided to go with clandestined.  It uses a C extension for the murmur3 consistent hashing algorithm which is pretty much necessary for performance requirements but I think its fine.

My test with 10,000 keys, this distribution looks like this:

![crc32](https://cloud.githubusercontent.com/assets/151924/8268809/63a3da04-1744-11e5-9f76-827776864ed3.png)
![ketama](https://cloud.githubusercontent.com/assets/151924/8268810/63a97338-1744-11e5-9c8c-996e27f47f21.png)
![murmur3](https://cloud.githubusercontent.com/assets/151924/8268811/63ab2caa-1744-11e5-8b0b-a993cea0fabd.png)

You can see using HRW (Rendezvous) balancing with a Murmur3 hash we are able to prevent any hotspots and evenly distribute the keys across all our servers.

I also checked how many keys would have to be redistributed if we dropped off on server and get the following:

```
CRC3 Redistribution 51.04
Murmur3 Redistribution 50.3
Ketama Redistribution 59.8
```

Murmur3 had the most evenly distributed keys and had to rebalance the lease amount.  Ketama (which is what pylibmc uses) has the worst amount of hotspots and has to redistribute the most.

duanhongyi:
Code LGTM.

PR #59, opened 2015-06-20
Add hashing client

sontek:
This is the beginning of how I would expect the hashing client to work.  It is built off of https://github.com/pinterest/pymemcache/pull/51 so that would need to be merged first.

It is an alternative to #44 and a fix for #52 

Things left to do:

- [x] Implement all client functions (set_many, get_many, incr, etc)
- [x] Implement failover detection if a client is detected as down
- [x] Write Integration Tests
- [x] Write Unit tests
- [x] Add proper logging
- [ ] Write detailed documentation

For failover support `HashClient` will support the following settings:

`retry_attempts` - After retry_attempts has been hit it will mark the server as dead and remove it from the rotation.
`retry_timeout` - It will wait this long in between retries
`dead_timeout` - After `retry_attempts` have been reached it will wait this long before attempting to add the server back into the pool.

Once a server is marked dead the keys that were allocated to it will be rebalanced across the nodes that are still up.

You can use this like this:

```python

from pymemcache.client.hash import HashClient
import time

client = HashClient([
    ('127.0.0.1', 11211),
    ('127.0.0.1', 11212)
])

while True:
    for key in ['foo', 'bar', 'baz']:
        client.set(key, 'bar')
        result = client.get(key)
        print(result)
    time.sleep(1)
```


sontek:
With babb8c0 I added support for failover.  This doesn't recover dead servers yet but it will remove misbehaving servers from the pool.

sontek:
8879147 brings dead servers back into rotation after the `dead_timeout`

sontek:
With  2470508 this is fairly feature complete.  There is a lot of polish I need to throw in like better logging / error checking and plenty more unit tests but its definitely working.

cgordon:
I did a quick first pass on the code, and it looks good! I'll take a closer look this week, probably on Wednesday (that's the first day I don't have wall-to-wall meetings). We use a memcache proxy internally, so we haven't needed this, but it's been high on my list of things to add for people that aren't using a proxy.

sontek:
@cgordon Sounds good, what proxy do you guys use?   We currently use `pylibmc` with `ketama` but have thought about using mcrouter or twemproxy instead

sontek:
@cgordon One thing I thought about with this is making `clandestined` an optional extra `pip install pymemcache[murmur3]` or something.  Then we can default to using crc32 which is built into the standard library.  crc32 isn't perfect but it works

pymemcache/client/base.py
```diff
@@ -684,7 +650,7 @@ def _fetch_cmd(self, name, keys, expect_cas):
                     result[key] = value
                 else:
                     raise MemcacheUnknownError(line[:32])
-        except Exception:
+        except Exception as e:
```
cgordon: Why do we need the "as e" here?

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+        key_prefix=b'',
+        max_pool_size=None,
+        lock_generator=None,
+        retry_attempts=2,
+        retry_timeout=1,
+        dead_timeout=60,
+        use_pooling=False,
+    ):
+        """
+        Args:
```
cgordon: Add a short description, just "Constructor." will do the job.

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+          servers: list(tuple(hostname, port))
+          serializer: optional class with ``serialize`` and ``deserialize``
```
cgordon: I agree with your argument that "serializer" and "deserializer" should be implemented as a class, but it's probably better to keep this consistent with the main Client class, so people can switch easily between them.

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+                  and ``remove_node``
+                  defaults to Rendezvous (HRW) hash.
+
+          use_pooling: use py:class:`.PooledClient` as the default underlying
+                       class. ``max_pool_size`` and ``lock_generator`` can
+                       be used with this. default: False
+
+          retry_attempts: Amount of times a client should be tried before it
+                          is marked dead and removed from the pool.
+          retry_timeout: Time in seconds that should pass between retry
```
cgordon: Can you mark this as "int" or "float", so it's more obvious whether fractional seconds are allowed. Same for the following.

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+        _check_key(key, self.key_prefix)
+        if len(self._dead_clients) > 0:
+            current_time = time.time()
+            ldc = self._last_dead_check_time
+            # we have dead clients and we have reached the
+            # timeout retry
+            if current_time - ldc > self.dead_timeout:
+                for server, dead_time in self._dead_clients.items():
+                    if current_time - dead_time > self.dead_timeout:
+                        print('bringing server back in rotation')
```
cgordon: Make this a log statement instead of a print to avoid contaminating stdout (or at least print to sys.err).

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+                        self.add_server(*server)
+                        self._last_dead_check_time = current_time
+
+        server = self.hasher.get_node(key)
+        print('got server %s' % server)
```
cgordon: Same.

pymemcache/client/hash.py
```diff
@@ -0,0 +1,282 @@
...
+                    # the sever as dead
+                    print('marking as dead')
+                    self.remove_server(*client.server)
+
+            result = func(*args, **kwargs)
+            return result
+
+        # Connecting to the server fail, we should enter
+        # retry mode
+        except socket.error:
```
cgordon: The fact that you aren't re-raising the exception anywhere in the "except" block means that failures are masked from callers of this method (and methods that call it). That might be alright, as long as it is carefully documented, but it means that this class has a very different contract than the standard Client, which is something we should think about pretty carefully.

cgordon:
@sontek The code looks great, I just have a few cosmetic comments and one comment about the overall contract of the methods. Let's talk about that a little bit when you have time.

Also, we're currently using mcrouter, and previously used twemproxy. I haven't been very involved in those decisions, but my understanding is that mcrouter is the better of the two.

pymemcache/client/base.py
sontek: Oops, added that during a debugging session, wasn't supposed to be included in the commit

pymemcache/client/hash.py
sontek: Yeah, I was trying to improve things for the future but I think instead of changing the contract during this PR I'll introduce a `v2` directory that changes the construct and raises deprecation warnings on the others in a different PR

pymemcache/client/hash.py
sontek: Yeah, the hard part is thats kind of the whole point of this client.  You really do want a graceful handling of failures so that nodes can be added/removed as problems are detected.  I specifically only used `socket.error` for now so that I wasn't masking other exceptions (for example, if you pass larger than 1mb data a server error would still be raised).  We could add the `ignore_exc` flag back and have them pass that before this graceful behavior kicks in?

I'm not really a fan of the ignore_exc in the other clients, I think they should raise exceptions

sontek:
@cgordon Addressed all of your concerns in a935261, what do you think about the overall contract?   I initially was trying to clean the contract up with this but I think we can do that slower with `v2` clients and deprecation warnings for the old ones.

I'm not completely against having the hashing client raise an exception but it would mean you would have to opt-in to graceful failover (unless we made it not raise exceptions by default)

sontek:
@cgordon The big question I have about this is if the murmur3/clandestined hash client should be the default.  clandestined uses `murmur3` C extension for performance, which makes this client no longer pure python.

What I think might be better is provide a less efficient/less consistent default using CRC32 which is pure python and provide an extras (client.ext.murmur3) that can be used like:

`pip install pymemcache[murmur3]` and then they will get the better client.  What do you think?  Is it fine to depend on the C extension in clandestined or should we make it optional?

sontek:
@cgordon I decided to go ahead and make a pure python3 version of this and make using the C dependency a documentation issue, here is the PR:

https://github.com/sontek/pymemcache/pull/1

I haven't merged it yet because I want to hear your feedback first.

pymemcache/client/hash.py
cgordon: I think I must be misunderstanding something about your comment or the code, so let me over-explain what I was trying to say so we can get on the same page.

Consider this code, using `HashClient`:

```
client = HashClient([("server1", 11211), ("server2", 11211), ("server3", 11211)])

val = client.get("my_key")
if val is None:
    val = my_service.some_method(some_args)
    client.set("my_key", val)
do_something(val)
```

This is a pretty typical interaction with a memcache client. Let's assume that HashClient has exactly the same contract for `get` and `set` (with `ignore_exc=False` by default). In that case, if "server1", "server2" and "server3" are all down, we would expect `client.get()` to raise an Exception. If the servers were available when `get` was called, but went down before `set` was called, then we would expect `set` to raise an Exception. If `ignore_exc=True`, then we would expect `get` to never raise (it would just return `None` when the servers were down), but `set` would still raise if the servers were unavailable.

What I'm pointing out is that the actual contract for `HashClient` has it always returning `None` from both `get` and `set` if the hashed server is down (even if some of the other servers are up, as we don't do retries on socket errors). That means `HashClient` acts like `ignore_exc=True` for both `get` and `set` methods at all times (and is not configurable). It's actually a little bit worse than that, as `_run_cmd()` always returns `None` when there is a socket error, and that breaks the contract of functions like `cas` and `get_many`, which either have a specific meaning for `None`, or should not ever return it (respectively). The implementation of `ignore_exc` in `Client` is careful to do the right thing for those cases.

There's no reason to choose between gracefully handling failures and supporting the `Client` contract. It is easy to do both. If you are uncomfortable with `ignore_exc`, then the simple way to do both is to add these lines:

```
...
except socket.error:
    exc_info = sys.exc_inf()
    ... your current failure logic ...
    raise exc_info[0], exc_info[1], exc_info[2]
```

If you would like to support the `ignore_exc` feature in `HashingClient` you'll have to wrap the last line of that in a conditional that checks the memcache call name and the value of ignore_exc and decides whether to return a value (we return `{}` in Client in that case) or raise.

FWIW, the `ignore_exc` feature is widely used (from what other users have told me), and is turned on by default at Pinterest, mostly because we do not want memcache server outages to cause site outages, and we don't want to wrap all calls to `get` in try/except clauses. I think it is worth supporting for `HashClient`, but I don't consider it a blocking change.

cgordon:
@sontek I agree with your last comment, that we should make the C dependency possible but optional. I took a look at the branch and it looks great to me.

cgordon:
@sontek Sorry for the long pause, was out of town for the 4th. Would you mind re-basing this change to pick up the documentation and test changes that we've already merged? The rest of the code looks great, but I just want to do one more pass through it before merging.

cgordon:
@sontek I'm ready to merge this change, but it needs to be merged with the documentation and testing changes from the last week. Can you take care of that?

sontek:
@cgordon Yeah, sorry I was busy last week but I'll clean it up in the next few days! :)

cgordon:
@sontek Awesome! Just double checking that I wasn't dropping the ball. Very excited about this change :)

sontek:
@cgordon I think its ready for review!  Want to take it for a spin on pinterest codebase to make sure I didn't break any backwards compatibility?

thedrow:
clandestined already doesn't compile murmur if on PyPy/PyPy3 and uses a pure python implementation of murmur instead following https://github.com/ewdurbin/clandestined-python/issues/8 so that's a non-issue.

sontek:
@thedrow Yeah, I think the important part is that `pymemcache` is pure python by default but people can opt-in to using clandestined with C dependencies if they want.   This allows us to use universal wheels and deploy with platforms without having to compile anything.

thedrow:
can we please add a build job that tests this both with clandestined and without?

thedrow:
can we please add a build job that tests this both with clandestined and without?

sontek:
@thedrow I'll document how to use it instead of just recommending it in the docs

cgordon:
LGTM. Thanks @sontek!

cgordon:
@sontek, separately, can you switch the readthedocs page to use the master repository so the docs are up to date? It looks like they point at your personal branch right now.

sontek:
@cgordon Done!  Do you have a readthedocs user? I can add you on there as well

cgordon:
@sontek I just created a user on readthedocs.org, I'm "charles.gordon".

PR #60, opened 2015-06-21
Added John Anderson (sontek) to contributors

sontek:


cgordon:
I'll push this up to pypi after we've added the hashing client and the testing changes.

PR #65, opened 2015-07-21
add default_kwargs to Client

duanhongyi:


cgordon:
LGTM.

@sontek FYI

sontek:
Whoops, would have loved to see a test attached to this fix so we don't screw that up again =)   I'll write one

PR #66, opened 2015-07-21
This adds a test to make sure kwargs are passed to client

sontek:


sontek:
@cgordon This test proves #65

PR #71, opened 2015-07-24
travis: check readme and flake8

methane:
Run flake8 and `setup.py check -rs` on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
   - memcached
 
 env:
+  - TOXENV=flake8
   - TOXENV=py26
   - TOXENV=py27
   - TOXENV=py33
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ commands =
     py.test {posargs}
 
 [testenv:flake8]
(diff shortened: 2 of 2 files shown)


cgordon:
Would this fix the problem with pypi's processing of the README.rst? When I run rst2html on the long description I am still seeing these errors:

```
<stdin>:29: (WARNING/2) Cannot analyze code. Pygments package not found.
<stdin>:35: (WARNING/2) Cannot analyze code. Pygments package not found.
````

methane:
Uh, Python 2.7.10 is required to check reST contains `code-block`.
See http://bugs.python.org/issue23063

cgordon:
Hmm, so we should just remove the code-block sections then? I don't think we should drop Python 2.6 support just to get slightly better documentation.

methane:
No. It's problem of only for `setup.py check`.
Travis runs it on Python 2.7.9 but the bug is fixed on 2.7.10.

methane:
Python 3.4.3 fixed the bug but Travis uses Python 3.4.2 too...
We can't check README on Travis for now.

PR #75, opened 2015-07-26
Enabled caching on Travis

thedrow:


.travis.yml
```diff
@@ -1,4 +1,7 @@
 sudo: false
+cache:
+  directories:
+    - $HOME/.cache/pip
```
methane: `cache: pip` is enough.
ref: http://docs.travis-ci.com/user/caching/#pip-cache

thedrow: It's not if you override the default install section. See http://docs.travis-ci.com/user/caching/#pip-cache

PR #76, opened 2015-08-12
make the hasher class pluggable in HashClient

slingamn:
I was doing some benchmarking and I needed this patch to be able to plug in ewdurbin/clandestined-python as the hash implementation.

In passing, there's a separate issue where pymemcache calls the relevant method `get_node`, but clandestined-python calls it `find_node`.

Thanks for your time.

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -15,7 +15,7 @@ class HashClient(object):
     def __init__(
         self,
         servers,
-        hasher=None,
+        hasher=RendezvousHash,
         serializer=None,
         deserializer=None,
         connect_timeout=None,
@@ -70,8 +70,7 @@ class HashClient(object):
         self._dead_clients = {}
         self._last_dead_check_time = time.time()
 
-        if hasher is None:
-            self.hasher = RendezvousHash()
+        self.hasher = hasher()
 
         self.default_kwargs = {
             'connect_timeout': connect_timeout,


cgordon:
@slingamn What is this change meant to fix? It doesn't appear to change the underlying functionality, just the way you pass information to the function. In addition, this change is not backwards compatible with existing clients, so we'd have to release it as a new major version.

slingamn:
Without this patch, the code only works when the default `None` value of `hasher` is not modified. When it is modified, you get a crash because `self.hasher` is never initialized:

      File "/nail/home/shivaram/benchmarks/memcache_bench.py", line 94, in <module>
        main()
      File "/nail/home/shivaram/benchmarks/memcache_bench.py", line 59, in main
        with client_manager() as client:
      File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
        return self.gen.next()
      File "/nail/home/shivaram/benchmarks/memcache_bench.py", line 41, in make_pymemcache_client
        yield pymemcache.client.hash.HashClient(servers, hasher=RendezvousHash)
      File "/nail/home/shivaram/benchmarks/pymemcache/client/hash.py", line 93, in __init__
        self.add_server(server, port)
      File "/nail/home/shivaram/benchmarks/pymemcache/client/hash.py", line 107, in add_server
        self.hasher.add_node(key)
    AttributeError: 'HashClient' object has no attribute 'hasher'

This change is not backwards compatible in the sense that it is a bug fix.

cgordon:
Got it, thanks for the fix!

sontek:
@slingamn Are you publishing your benchmarks anywhere?

slingamn:
I'm not planning on a formal writeup, but I could share the numbers here.

PR #79, opened 2015-08-28
Add default_noreply to Client

### LINKED ISSUE #78: Support default parameter for 'noreply'
When integrating this with Django I found it annoying that `noreply` was set, but I want to disable it. The Django interface is for generic k/v caches and thus doesn't support adding arguments - thus I had to subclass client and for *each* method pass `noreply=False` to the `super()` call. Would be nice if the `Client` classes by default did `noreply=None` and then `if noreply is None: noreply = self.default_noreply` so at instantiation time you could decide whether you wanted to `noreply` or not.

adamchainz:
Resolves #78 

Have only added to the base `Client` at this time, will extend to other classes if accepted as a feature suggestion.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -161,7 +161,8 @@ class Client(object):
                  no_delay=False,
                  ignore_exc=False,
                  socket_module=socket,
-                 key_prefix=b''):
+                 key_prefix=b'',
+                 default_noreply=True):
         """
         Constructor.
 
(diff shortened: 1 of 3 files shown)


cgordon:
LGTM, thanks!

cgordon:
I have a few things to release to pypi, and will have that done by Wednesday evening. Sorry for the delay, just a busy week.

adamchainz:
Cool - I'll add `default_noreply` to all the other `Client` classes in another PR coming soon.

Issue #77, opened 2015-08-19
KeyError when no connections are available using HashClient


renstrom:
When the `HashClient` runs out connections it will start raising `KeyError` (even with `ignore_exc=True`). Reproduce with this:

```python
>>> from pymemcache.client.hash import HashClient
>>> client = HashClient([('localhost', 11211)], use_pooling=True, ignore_exc=True,
...                     timeout=2, connect_timeout=2)
>>> client.set('foo', 'bar')
True
>>> client.get('foo')
'bar'
>>> # Stop Memcached server
...
>>> client.get('foo')
>>> client.get('foo')
>>> client.get('foo')
>>> client.get('foo')
>>> client.get('foo')
>>> client.get('foo')
>>> client.get('foo')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Peter/.vim/venv/lib/python2.7/site-packages/pymemcache/client/hash.py", line 226, in get
    return self._run_cmd('get', key, None, *args, **kwargs)
  File "/Users/Peter/.vim/venv/lib/python2.7/site-packages/pymemcache/client/hash.py", line 214, in _run_cmd
    client = self._get_client(key)
  File "/Users/Peter/.vim/venv/lib/python2.7/site-packages/pymemcache/client/hash.py", line 134, in _get_client
    client = self.clients[server]
KeyError: None
```

cgordon:
@sontek 

I took a look at the code. This is caused by the _get_client method in HashClient returning None when there are no live nodes. Fixing that is going to require a little bit of code restructuring, since there isn't a good way to return a correct value from the place the error occurs. I'll have time for this next week, adding John Anderson in case he has time sooner.

sontek:
Thanks for poking me, I didn't notice this error report. I'll fix it up tonight

sontek:
@cgordon @renstrom Sorry I got busy the other night but #81 will make it graceful when ignore exceptions are set.

cgordon:
Will close this when it is deployed to pypi (this weekend).

PR #81, opened 2015-09-06
hashclient ignores exceptions when no servers are found

sontek:
This fixes the hashclient when `ignore_exc` is set to `True`.   I will make a separate PR on top of this that will get no servers found put into the retry cycle.  So if all servers are down but we haven't waited our defined time it wont blow up either.  Just need to think about that use case a little more :)

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -130,6 +130,12 @@ class HashClient(object):
                         self._last_dead_check_time = current_time
 
         server = self.hasher.get_node(key)
+        # We've ran out of servers to try
+        if server is None:
+            if self.ignore_exc is True:
+                return
+            raise Exception('All servers seem to be down right now')
+
         client = self.clients[server]
         return client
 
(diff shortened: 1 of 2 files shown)


sontek:
This fixes #77

pymemcache/client/hash.py
```diff
@@ -130,6 +130,12 @@ def _get_client(self, key):
                         self._last_dead_check_time = current_time
 
         server = self.hasher.get_node(key)
+        # We've ran out of servers to try
+        if server is None:
+            if self.ignore_exc is True:
```
renstrom:
```diff
-            if self.ignore_exc is True:
+            if self.ignore_exc:
```

sontek: @renstrom I always do that unless there is a reason to support all falsy operations, i.e do you think this should work:

```python
self.ignore_exc = 1
```

renstrom: Right, fair enough.

cgordon:
I'll get this out to pypi over the weekend, thanks for the quick fix!

PR #86, opened 2015-09-28
Add myself to credits.

ewdurbin:
with the significant amount of `python-clandestined` source code copied into the `HashClient`, I would appreciate accreditation somewhere high-level than in the depths of source code.


cgordon:
Definitely, sorry for the oversight!

ewdurbin:
Thank you!

PR #89, opened 2015-10-21
Fixed bugs caused by setting retry attempts to 0 on HashClient

Vhab:
This pull fixes 2 small issues.

The first is an exception thrown when `retry_attempts` to 0 instead of -1 to disable it. It would throw a key error due to `self._failed_clients` not being populated.

The second issue is an exception regarding the use of % while creating output log and `client.server` being a tuple instead of a string.

Included testcase fails on current master and succeeds on this branch.


cgordon:
Thank you @Vhab! I'll add you to the acknowledgments and get this out to pypi later this week.

Issue #92, opened 2016-03-17
Unpack error with Ubuntu's (14.04) release of memcached


RobertDeRose:
on line 702 of pymemcache/client/base.py
```python
701                 elif name == b'stats' and line.startswith(b'STAT'):
702                     _, key, value = line.split()
703                     result[key] = value
```

line 702 throws an exception with the default memcached from Ubuntu.

The stats command expects all output to be exactly three values separated by spaces, however, this is the result for the stats commands on Ubuntu 14.04

b'STAT version 1.4.14 (Ubuntu)'

Connecting to memcached through telnet and issuing the `stats` command gives you:
```bash
STAT pid 993
STAT uptime 11057
STAT time 1458238357
STAT version 1.4.14 (Ubuntu)
STAT libevent 2.0.21-stable
STAT pointer_size 64
STAT rusage_user 1.037698
STAT rusage_system 0.325248
.
.   <- cut for brevity ->
.
STAT reclaimed 2
```

As you can see, all output does follow this pattern, except for the `version` line.
This could be submitted to Ubuntu upstream as a bug since they appear to have broken the pattern, but I'm not sure it's documents (haven't researched it) as being required to fix this pattern.

PR #93, opened 2016-03-17
Handle stats not conforming to the triplet format

RobertDeRose:
Addresses issue #92

In Ubuntu's (14.04) version of memecached packaged with the OS the
version STAT line contains a fourth value due to a space. This breaks
the expected triplet pattern and incorrectly caused an exception.


cgordon:
Looks good, I'll get this out to pypi over the weekend. Thanks!

RobertDeRose:
:fireworks:  Thanks!

Issue #94, opened 2016-04-07
stats value for evictions is boolean when it should be numeric?


amcrn:
per https://github.com/memcached/memcached/blob/b1debc4c96134e9014a929c32d6990cb5d66a22c/doc/protocol.txt#L562-L563 the value for `evictions` from a `stats` command is numeric.

but pymemcache is treating it as on/off: https://github.com/pinterest/pymemcache/blob/aac032419e94b5f71e763c130d54d4b6f36caa17/pymemcache/client/base.py#L56

this seems odd, because if memcached returns `evictions` as `55`, pymemcache is returning this as `False`.

i'd send a pull request, but it seems like this might have done for a reason that's escaping me at the moment ;) let me know.

cgordon:
That looks like a bug, thanks for the report! If you have time to put together a pull request that would be great. Otherwise I'll add it to my to-do list :)

rsanr:
@cgordon I work with @amcrn and will issue a pull request soon.

PR #96, opened 2016-04-15
Fixed stats evictions value conversion

rsanr:
Addresses issue #94 

Fixed evictions value conversion in base.py.  
- evictions stat value was incorrectly being compared to “b’on’” previously, which resulted in a boolean. The fix makes the conversion default to int, which is in line with other numeric stats values.

Augmented test_stats_conversions() test case in test_client.py to test evictions value conversion

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -53,7 +53,6 @@ STAT_TYPES = {
 
     # Settings stats
     b'inter': six.binary_type,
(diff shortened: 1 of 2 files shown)


cgordon:
Can you add a test to make sure this doesn't regress later?

rsanr:
I augmented test_stats_conversions() test in test_client.py to test evictions value conversion

cgordon:
Thanks!

PR #101, opened 2016-06-25
update comparison to pylibmc

slingamn:
pylibmc v1.5.1 includes a serialization API: http://sendapatch.se/projects/pylibmc/reference.html#pylibmc.Client.serialize


cgordon:
Thank you!

PR #104, opened 2016-07-27
Syntax and build

nichochar:
Running `tox` left the project with some syntax and build errors. This cleans it up a little so that cloning and running `tox` should run fine


jparise:
LGTM

PR #105, opened 2016-07-27
Security sanitize key 

### LINKED ISSUE #99: Security issue in input handling
Key is not sanitized and something like \r\nflush_all\r\n in it could destroy all values and maybe more, if it comes from any user input.

nichochar:
Fixes #99

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -72,7 +72,7 @@ def _check_key(key, key_prefix=b''):
         except UnicodeEncodeError:
             raise MemcacheIllegalInputError("No ascii key: %r" % (key,))
     key = key_prefix + key
-    if b' ' in key:
+    if b' ' in key or b'\r\n' in key:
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -72,7 +72,7 @@ def _check_key(key, key_prefix=b''):
         except UnicodeEncodeError:
             raise MemcacheIllegalInputError("No ascii key: %r" % (key,))
     key = key_prefix + key
-    if b' ' in key:
+    if b' ' in key or b'\r\n' in key:
```
jparise: Are we only worried about `\r\n` (and `\r` and `\n` are individually acceptable)?

nichochar: After talking to folks in the memcached IRC channel, this is sufficient for security issues. But I just tested manually using both of them, and `\n` breaks the client for 1 call.. 

I will add a fix for this in this PR

nichochar: Fixed.

jparise:
👍

PR #107, opened 2016-08-01
Added default parameter support for "get" and "gets" methods in Client

Morreski:
I thought it would be a nice feature to be able to do:

```python
foo = client.get('foo', default=b'bar').decode()
```

instead of:

```python
foo = client.get('foo')
if foo is None:
    foo = b'foo'
foo.decode()
```

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -377,17 +377,18 @@ class Client(object):
         """
         return self._store_cmd(b'cas', key, expire, noreply, value, cas)
 
-    def get(self, key):
+    def get(self, key, default=None):
         """
         The memcached "get" command, but only for one key, as a convenience.
 
         Args:
           key: str, see class docs for details.
+          default: value that will be returned if the key was not found.
 
         Returns:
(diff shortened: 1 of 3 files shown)


nichochar:
This looks good to me, I think it's a nice added feature.
Thanks for the tests, and the documentation.
Unless any other maintainers object, I will merge this in by EOW

cgordon:
👍

pymemcache/client/base.py
```diff
@@ -408,17 +409,21 @@ def get_many(self, keys):
...
 
         Args:
           key: str, see class docs for details.
+          default: value that will be returned if the key was not found.
+          cas_default: same behaviour as default argument.
 
         Returns:
-          A tuple of (key, cas), or (None, None) if the key was not found.
+          A tuple of (key, cas)
+          or (default, cas_defaults) if the key was not found.
```
Morreski: Oops just spotted the typo here: cas_defaults shouldn't have the trailing S

Issue #102, opened 2016-07-18
Fix Documentation


nichochar:
The documentation written inline in the code isn't viewable in readthedocs. The whole sphinx documentation building needs a fix.

PR #108, opened 2016-08-03
Docs

nichochar:
Making the inline documentation appear in the "modules" section on ReadTheDocs

This _should_ fix https://github.com/pinterest/pymemcache/issues/102 but it's hard to tell since read the docs only builds the master branch.


docs/Makefile
```diff
@@ -0,0 +1,225 @@
...
+	@echo "  texinfo    to make Texinfo files"
+	@echo "  info       to make Texinfo files and run them through makeinfo"
+	@echo "  gettext    to make PO message catalogs"
+	@echo "  changes    to make an overview of all changed/added/deprecated items"
+	@echo "  xml        to make Docutils-native XML files"
+	@echo "  pseudoxml  to make pseudoxml-XML files for display purposes"
+	@echo "  linkcheck  to check all external links for integrity"
+	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
+	@echo "  coverage   to run coverage check of the documentation (if enabled)"
+	@echo "  dummy      to check syntax errors of document sources"
```
jparise: Would this dummy target be a useful thing to run as part of Travis' build?

jparise:
LGTM

docs/Makefile
nichochar: I will investigate, this could indeed be useful, but I adding a coverage test will be in a separate PR.

PR #109, opened 2016-08-03
Automate doc building

nichochar:
This will build the docs automatically when they go through the readthedocs process, as explained here
https://github.com/rtfd/readthedocs.org/issues/1139


jparise:
LGTM

PR #110, opened 2016-08-16
Improve MemcacheIllegalInputError messages

adamchainz:
* Quote the key so it's easier to realize e.g. trailing spaces
* Point out newlines are bad as well as spaces
* Capitalize ASCII


nichochar:
LGTM, merging.

PR #111, opened 2016-08-16
Improve stats parsing

adamchainz:
Create one function for each extra parseable type to make it clear there are different types in the stats response, and to stop wasting memory on making several `lambda`s that all do the same thing.

Also improve test coverage of the parsing.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -40,25 +40,41 @@ VALID_STORE_RESULTS = {
 
 # Some of the values returned by the "stats" command
 # need mapping into native Python types
+def _parse_bool_int(value):
+    return int(value) != 0
+
+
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -40,25 +40,41 @@
 
 # Some of the values returned by the "stats" command
 # need mapping into native Python types
+def _parse_bool_int(value):
+    return int(value) != 0
+
+
+def _parse_bool_string(value):
```
nichochar: Can we name this something slightly more descriptive, like `_parse_bool_string_is_yes` ?

nichochar:
Could we improve the naming a little? A part from that, this is good 👍 will merge once names are updated.

PR #112, opened 2016-08-16
Lint on Python 3 too

adamchainz:
* Add linting on Python 3 too - normal environment failed for me initially since my `tox` is installed with Python 3 and so `flake8` failed on the syntax errors in `serde`
* Fix `serde` to be importable in Python 3
* Add some tests for `serde`

### DIFF
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -14,11 +14,7 @@
 
 import logging
 import pickle
-
-try:
-    from cStringIO import StringIO
-except ImportError:
-    from StringIO import StringIO
+from six.moves import cStringIO as StringIO
 
 try:
     long_type = long  # noqa
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -14,7 +14,7 @@
 
 import logging
 import pickle
-from six.moves import cStringIO as StringIO
+from io import BytesIO
 
 try:
     long_type = long  # noqa
@@ -40,7 +40,7 @@ def python_memcache_serializer(key, value):
         value = "%d" % value
     else:
         flags |= FLAG_PICKLE
-        output = StringIO()
+        output = BytesIO()
         pickler = pickle.Pickler(output, 0)
         pickler.dump(value)
         value = output.getvalue()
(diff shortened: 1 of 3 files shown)


pymemcache/serde.py
```diff
@@ -14,11 +14,12 @@
...
 import logging
 import pickle
+from io import BytesIO
 
 try:
-    from cStringIO import StringIO
-except ImportError:
-    from StringIO import StringIO
+    long_type = long  # noqa
+except NameError:
```
nichochar: When does this get raised?

pymemcache/test/test_serde.py
```diff
@@ -0,0 +1,24 @@
+from unittest import TestCase
+
+from pymemcache.serde import (python_memcache_serializer,
+                              python_memcache_deserializer)
+
+
+class TestSerde(TestCase):
```
nichochar: Thank you for adding this

nichochar:
This looks good to me. @jparise , since you have some experience with python3 compatibility, could you please also give this a quick check? Then I'll merge.

pymemcache/serde.py
```diff
@@ -55,11 +56,11 @@ def python_memcache_deserializer(key, value, flags):
         return int(value)
 
     if flags & FLAG_LONG:
-        return long(value)
+        return long_type(value)
```
jparise: Can't `long_type` be `None` here?

nichochar: Not from what I can tell, it can only be set as `FLAG_LONG` line 39, and that only happens if `long_type` is explicitely not `None`

adamchainz: `long` does not exist in Python 3

In fact even calling `long(x)` is unnecessary on Python 2 since `int`s automatically become `long`s at a certain size.

I'd say drop the `long` flag altogether but obviously that's impossible now since people have serialized and stored data with this code.

pymemcache/serde.py
adamchainz: `long` does not exist on Python 3


```sh
$ python3
Python 3.5.2 (default, Jul 24 2016, 10:28:00)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> long
NameError: name 'long' is not defined
>>>
```

nichochar: 👍

PR #115, opened 2016-09-19
Fix MockMemcacheClient serialization/deserialization

suhailpatel:
There is a slight issue with the `MockMemcacheClient` where it's not using the serializer/deserializer properly and extracting out values and flags. 

This PR fixes and adds a few tests around this behaviour


nichochar:
This looks good to me, thanks for the cleanup.

PR #116, opened 2016-10-03
Use cPickle if available on Python 2

suhailpatel:
Using cPickle can result in big speed boosts according to the [Python docs on `pickle`](https://docs.python.org/2/library/pickle.html#relationship-to-other-python-modules)

> The pickle module has an optimized cousin called the cPickle module. As its name implies, cPickle is written in C, so it can be up to 1000 times faster than pickle

For Python 2, you need to import `cPickle` explicitly. In Python 3, `import pickle` chooses `cPickle` automatically if it's available.

### DIFF
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -14,15 +14,7 @@
 
 import logging
 from io import BytesIO
-import six
-
-if six.PY2:
-    try:
-        import cPickle as pickle
-    except ImportError:
-        import pickle
-else:
-    import pickle
+from six.moves import cPickle as pickle
 
 try:
     long_type = long  # noqa
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -12,8 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
+
 import logging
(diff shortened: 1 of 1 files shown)


jparise:
Because we're already using `six`, I believe you can also just import the "moved" symbol:

```python
from six.moves import cPickle as pickle
```

nichochar:
@suhailpatel thanks for the PR! Can we use `six.moves` as suggested by @jparise ?
A part from that this looks good, it would be nice to have a benchmark of any kind of speed increase. Is this something you could maybe also look into ?

suhailpatel:
Addressed comments. Thanks for pointing out that `six` already handles this

Regarding benchmarking, it is quite a speedup on Python 2. I wrote a simple benchmark script that pickled and unpickled an object 100000 times (link to script: https://gist.github.com/suhailpatel/968f56080656b8521ee87dbe9f7ddea3) and the results on my machine were as follows

Before change:
```
Time to serialize 100000 times: 10.4943289757
Time to deserialize 100000 times: 9.19699287415
```

After change:
```
Time to serialize 100000 times: 2.56918907166
Time to deserialize 100000 times: 2.2167420387
```

nichochar:
Thanks for the PR, making the changes, and the benchmark! LGTM

PR #118, opened 2016-10-03
Fix #117, illegal unicode character in middle of key (or value now)

### LINKED ISSUE #117: Unicode character in key not detected if python2 str
User tried to use key `'rpgv:https://www.pacificsales.com/brands/viking?ref=166_725&loc=soc_\xb1901520_%c3%abuy!'` and no `MemcacheIllegalInputError` was raised, which causes obsure bug with stack trace below

```
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/services/utils/task_flow.py", line 165, in _run
return self.run()
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/services/utils/task_flow.py", line 160, in run
return self.f(*self.args, **self.kwargs)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/common/utils/decorators.py", line 556, in __get__
value = self.func(obj)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/api/serializers/contexts.py", line 1040, in rich_summary_map
return core.RichPinGridData.manager.get_many(ids)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/core/managers/rich_pin/rich_pin_view_manager.py", line 143, in get_many
return _get_many(model_ids)
File "/mnt/virtualenv/local/lib/python2.7/site-packages/pinstatsd/statsd.py", line 77, in decorated_function
return f(*args, **kwargs)
File "/mnt/virtualenv/local/lib/python2.7/site-packages/pinstatsd/statsd.py", line 98, in decorated_function
return_value = func(*args, **kwargs)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/core/managers/rich_pin/rich_pin_view_manager.py", line 138, in _get_many
models = super(FilteredRichPinViewManager, self).get_many(model_ids)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/core/managers/hbase_managers.py", line 572, in get_many
many_as_dicts = self.get_many_as_dicts(model_ids)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/core/managers/hbase_managers.py", line 552, in get_many_as_dicts
fetched_dicts = _get_many_as_dicts(self, model_ids)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/data_clients/mc_objects.py", line 154, in fn
lzero_missed_keys) or {}
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/data_clients/memcache.py", line 222, in wrap_f
return f(self, *args, **kwargs)
File "/mnt/builds/qB9LY9wQQXa4xZTJHZEv8A_d2eb341/data_clients/memcache.py", line 451, in get_many
result = mc.get_many(keys)
File "/mnt/virtualenv/local/lib/python2.7/site-packages/pymemcache/client/base.py", line 407, in get_many
return self._fetch_cmd(b'get', keys, False)
File "/mnt/virtualenv/local/lib/python2.7/site-packages/pymemcache/client/base.py", line 689, in _fetch_cmd
key = checked_keys[key]
KeyError: 'rpgv:https://www.pacificsales.com/brands/viking?ref=166_725&loc=soc_\xb1901520_%c3%abuy!'
```

nichochar:
See #117 
This fixes an error where one lonely unicode character can _hide_ in a key or value. We need to make sure to raise the illegal input when this happens.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -82,7 +82,8 @@ STAT_TYPES = {
 
 def _check_key(key, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, six.text_type) or isinstance(key, six.string_types):
+    allowed_str_types = (six.text_type, six.string_types)
+    if isinstance(key, allowed_str_types):
         try:
             key = key.encode('ascii')
         except (UnicodeEncodeError, UnicodeDecodeError):
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -82,10 +82,10 @@ STAT_TYPES = {
 
 def _check_key(key, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, six.text_type):
+    if isinstance(key, six.text_type) or isinstance(key, six.string_types):
         try:
             key = key.encode('ascii')
-        except UnicodeEncodeError:
+        except (UnicodeEncodeError, UnicodeDecodeError):
             raise MemcacheIllegalInputError("Non-ASCII key: '%r'" % (key,))
     key = key_prefix + key
     if b' ' in key or b'\n' in key:
--- a/pymemcache/test/test_client.py
+++ b/pymemcache/test/test_client.py
@@ -92,6 +92,15 @@ class ClientTestMixin(object):
         with pytest.raises(MemcacheIllegalInputError):
             _set()
 
+    def test_set_unicode_char_in_middle_of_key(self):
+        client = self.make_client([b'STORED\r\n'])
+
+        def _set():
+            client.set('helloworld_\xb1901520_%c3', b'value', noreply=False)
+
+        with pytest.raises(MemcacheIllegalInputError):
+            _set()
+
     def test_set_unicode_value(self):
         client = self.make_client([b''])
 
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ env/
 .pip
 .pypirc
 coverage.xml
+.python-version
 \#*\#
 docs/_build
 docs/apidoc/pymemcache.test.rst
--- a/pymemcache/test/utils.py
+++ b/pymemcache/test/utils.py
@@ -43,6 +43,11 @@ class MockMemcacheClient(object):
     def get(self, key, default=None):
         if isinstance(key, six.text_type):
             raise MemcacheIllegalInputError(key)
+        if isinstance(key, six.string_types):
+            try:
+                key = key.encode('ascii')
+            except (UnicodeEncodeError, UnicodeDecodeError):
+                raise MemcacheIllegalInputError
 
         if key not in self._contents:
             return default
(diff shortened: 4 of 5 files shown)


pymemcache/client/base.py
```diff
@@ -82,10 +82,10 @@ def _parse_hex(value):
 
 def _check_key(key, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, six.text_type):
+    if isinstance(key, six.text_type) or isinstance(key, six.string_types):
```
jparise: We might want to introduce a top-level variable that defines the set of allowed key types so that this can become a single `isinstance` check:

```python
key_types = six.string_types + (six.text_types,) # or a similar concatenation

if isinstance(key, key_types):
```

pymemcache/test/test_client.py
```diff
@@ -92,6 +92,15 @@ def _set():
...
             _set()
 
+    def test_set_unicode_char_in_middle_of_key(self):
+        client = self.make_client([b'STORED\r\n'])
+
+        def _set():
+            client.set('helloworld_\xb1901520_%c3', b'value', noreply=False)
+
+        with pytest.raises(MemcacheIllegalInputError):
+            _set()
```
jparise: Why `_set()` instead of just dropping the `client.set(...)` right here? Just copying the pattern used above?

pymemcache/client/base.py
nichochar: is this a possible thing? I like it, it's more elegant. It's slightly less obvious to name than `allowed_key_types`, since it's only dealing with strings. I believe you can have booleans and bytes (and then there's the whole python2/3 thing for bytes)

pymemcache/test/test_client.py
nichochar: Yes, I was copying. I have no good reason, do you think I should change them? I don't know if it belongs in this particular PR

pymemcache/client/base.py
jparise: Yes, you can pass `isinstance` a sequence of types to check.

pymemcache/client/base.py
```diff
@@ -82,10 +82,11 @@ def _parse_hex(value):
 
 def _check_key(key, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, six.text_type):
+    allowed_str_types = (six.text_type, six.string_types)
```
jparise: This could be global so we don't need to rebuilt the tuple each time this function runs.

pymemcache/test/test_client.py
jparise: Separately is fine.

PR #125, opened 2016-10-28
[fix #123] PooledClient now use default_noreply arg set on init

### LINKED ISSUE #123: `PooledClient` doesn't use `default_noreply` arg
This arg is passed to the underlying Client:
https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L881

But not honoured when the request is sent, as the `noreply` arg is explicitly set to `True`:
https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L887

kols:


kols:
@nichochar please review

nichochar:
This is great, thank you for the patch and tests @kols, it LGTM

PR #128, opened 2016-10-31
Add optional support for unicode keys

jogo:
memcached's ASCII protocol supports unicode keys, so lets support them
as well. Since using unicode keys for memcache is uncommon and to
preserve the previous behavior disable support by default.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -123,3 +123,4 @@ Credits
 * `Ernest W. Durbin III <https://github.com/ewdurbin>`_
 * `Remco van Oosterhout <https://github.com/Vhab>`_
 * `Nicholas Charriere <https://github.com/nichochar>`_
+* `Joe Gordon <https://github.com/jogo>`_
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -63,8 +63,11 @@ This client implements the ASCII protocol of memcached. This means keys should n
 contain any of the following illegal characters:
 > Keys cannot have spaces, new lines, carriage returns, or null characters.
 We suggest that if you have unicode characters, or long keys, you use an effective
-hashing mechanism before calling this client. At Pinterest, we have found that murmur3 hash is a
-great candidate for this.
+hashing mechanism before calling this client. At Pinterest, we have found that
+murmur3 hash is a great candidate for this. Alternatively you can
+set `allow_unicode_keys` to support unicode keys, but beware of
+what unicode encoding you use to make sure multiple clients can find the
+same key.
 
 
 Best Practices
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -81,9 +81,12 @@ STAT_TYPES = {
 # Common helper functions.
 
 
-def _check_key(key, key_prefix=b''):
+def _check_key(key, allow_unicode_keys, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, VALID_STRING_TYPES):
+    if allow_unicode_keys:
+        if isinstance(key, six.text_type):
+            key = key.encode('utf8')
+    elif isinstance(key, VALID_STRING_TYPES):
         try:
             key = key.encode('ascii')
         except (UnicodeEncodeError, UnicodeDecodeError):
@@ -177,7 +180,8 @@ class Client(object):
                  ignore_exc=False,
                  socket_module=socket,
                  key_prefix=b'',
-                 default_noreply=True):
+                 default_noreply=True,
+                 allow_unicode_keys=False):
         """
         Constructor.
 
@@ -203,6 +207,7 @@ class Client(object):
           default_noreply: bool, the default value for 'noreply' as passed to
             store commands (except from cas, incr, and decr, which default to
             False).
+          allow_unicode_keys: bool, support unicode (utf8) keys
 
         Notes:
           The constructor does not make a connection to memcached. The first
@@ -223,10 +228,11 @@ class Client(object):
             raise TypeError("key_prefix should be bytes.")
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
+        self.allow_unicode_keys = allow_unicode_keys
 
     def check_key(self, key):
         """Checks key and add key_prefix."""
-        return _check_key(key, key_prefix=self.key_prefix)
+        return _check_key(key, allow_unicode_keys=self.allow_unicode_keys, key_prefix=self.key_prefix)
 
     def _connect(self):
         sock = self.socket_module.socket(self.socket_module.AF_INET,
@@ -841,7 +847,8 @@ class PooledClient(object):
                  key_prefix=b'',
                  max_pool_size=None,
                  lock_generator=None,
-                 default_noreply=True):
+                 default_noreply=True,
+                 allow_unicode_keys=False):
         self.server = server
         self.serializer = serializer
         self.deserializer = deserializer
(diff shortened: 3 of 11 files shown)


pymemcache/client/base.py
```diff
@@ -81,13 +81,26 @@ def _parse_hex(value):
 # Common helper functions.
 
 
-def _check_key(key, key_prefix=b''):
+def _bytearray(key):
```
methane: Using bytearray to bypass `if isinstance(key, VALID_STRING_TYPES):` doesn't make sense.

pymemcache/client/base.py
```diff
@@ -203,6 +217,7 @@ def __init__(self,
           default_noreply: bool, the default value for 'noreply' as passed to
             store commands (except from cas, incr, and decr, which default to
             False).
+          unicode_key_support: bool, support unicode (utf8) keys
```
methane: I prefer "allow_noascii_key"

pymemcache/client/base.py
```diff
@@ -683,8 +699,13 @@ def _raise_errors(self, line, name):
             raise MemcacheServerError(error)
 
     def _fetch_cmd(self, name, keys, expect_cas):
-        checked_keys = dict((self.check_key(k), k) for k in keys)
-        cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
+        checked_keys = dict((tuple(self.check_key(k)), k) for k in keys)
```
methane: I don't understand why `tuple` here.

pymemcache/client/base.py
```diff
@@ -683,8 +699,13 @@ def _raise_errors(self, line, name):
...
+        concat = b""
+        for i, key in enumerate(checked_keys):
+            if i > 0:
+                concat += b" "
+            concat += _bytearray(key)
```
methane: This is unidiomatic and inefficient code.

jparise:
I agree we're at the point where our tests should also run against a real memcached instance. This is easy to run locally, and the Travis environment can be configured with a memcached service too:

```yaml
services:
  - memcached
```

jogo:
New revision coming shortly, not sure why I thought I needed bytearray.

pymemcache/client/base.py
nichochar: I also prefer that, maybe `allow_nonASCII_keys` ?

pymemcache/client/base.py
```diff
@@ -864,7 +872,7 @@ def __init__(self,
 
     def check_key(self, key):
         """Checks key and add key_prefix."""
-        return _check_key(key, key_prefix=self.key_prefix)
+        return _check_key(key, unicode_key_support=self.unicode_key_support, key_prefix=self.key_prefix, )
```
nichochar: Maybe trim this trailing comma since this convention has not been kept in the rest of the repo

pymemcache/client/rendezvous.py
```diff
@@ -36,7 +36,7 @@ def get_node(self, key):
 
         for node in self.nodes:
             score = self.hash_function(
-                "%s-%s" % (str(node), str(key)))
+                "%s-%s" % (node, key))
```
nichochar: Why did you make this change? This could be used to cast ints

docs/getting_started.rst
```diff
@@ -63,8 +63,11 @@ This client implements the ASCII protocol of memcached. This means keys should n
 contain any of the following illegal characters:
 > Keys cannot have spaces, new lines, carriage returns, or null characters.
 We suggest that if you have unicode characters, or long keys, you use an effective
-hashing mechanism before calling this client. At Pinterest, we have found that murmur3 hash is a
-great candidate for this.
+hashing mechanism before calling this client. At Pinterest, we have found that
+murmur3 hash is a great candidate for this. Alternatively you can
+set `unicode_key_support` to support unicode keys, just be careful about
```
nichochar: minor: maybe s/"just be careful about"/"but beware of"

pymemcache/client/base.py
```diff
@@ -81,9 +81,12 @@ def _parse_hex(value):
...
 
 
-def _check_key(key, key_prefix=b''):
+def _check_key(key, unicode_key_support, key_prefix=b''):
     """Checks key and add key_prefix."""
-    if isinstance(key, VALID_STRING_TYPES):
+    if unicode_key_support:
+        if isinstance(key, six.text_type):
+            key = key.encode('utf8')
+    elif isinstance(key, VALID_STRING_TYPES):
```
nichochar: nice syntax with `elif` here.

nichochar:
@methane @jparise we already do test against memcache, see our [travis file](https://github.com/pinterest/pymemcache/blob/master/.travis.yml#L8).

@methane I agree those test strings would be nice, would you like to make a pull request, or maybe create an issue?

Thanks!

pymemcache/test/test_integration.py
```diff
@@ -23,27 +24,47 @@
...
-    assert result == b'value'
-
-    client.set(b'key2', b'value2', noreply=True)
-    result = client.get(b'key2')
-    assert result == b'value2'
 
-    result = client.get_many([b'key', b'key2'])
-    assert result == {b'key': b'value', b'key2': b'value2'}
+@pytest.mark.integration()
+def test_get_set_unicode_key(client_class, host, port, socket_module):
```
jogo: @methane this is the unicode integration test you requested.

pymemcache/client/rendezvous.py
nichochar: Oh, you're right actually, the casting happens automatically with the string substitution, I missed that. Thanks

pymemcache/client/base.py
```diff
@@ -223,10 +228,11 @@ def __init__(self,
             raise TypeError("key_prefix should be bytes.")
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
+        self.allow_unicode_keys = allow_unicode_keys
 
     def check_key(self, key):
         """Checks key and add key_prefix."""
-        return _check_key(key, key_prefix=self.key_prefix)
+        return _check_key(key, allow_unicode_keys=self.allow_unicode_keys, key_prefix=self.key_prefix)
```
methane: This line looks bit longer than usual.
How about stopping keyword argument?

pymemcache/client/base.py
```diff
@@ -864,7 +872,7 @@ def __init__(self,
...
```
methane: same here.

pymemcache/client/hash.py
```diff
@@ -113,7 +116,7 @@ def remove_server(self, server, port):
         self.hasher.remove_node(key)
 
     def _get_client(self, key):
-        _check_key(key, self.key_prefix)
+        _check_key(key, self.allow_unicode_keys, self.key_prefix)
```
methane: Same here.

methane:
LGTM

README.rst
```diff
@@ -123,3 +123,4 @@ Credits
 * `Ernest W. Durbin III <https://github.com/ewdurbin>`_
 * `Remco van Oosterhout <https://github.com/Vhab>`_
 * `Nicholas Charriere <https://github.com/nichochar>`_
+* `Joe Gordon <https://github.com/jogo>`_
```
nichochar: Very happy to have a new contributor! 🙌 thanks @jogo

pymemcache/client/base.py
nichochar: I realized that we don't have an explicit style guidelines in the project. I will add a ticket to add one.
My personal preference is 120 lines, and this fits within that limit, so I am OK with it.

PR #131, opened 2016-11-15
Python 3 serialization fixes, serialization improvements

tomkins:
I was looking at trying to use/resurrect django-pymemcache, however in trying to update it and get tests working on it - pymemcache serialization seems to have some issues with byte/unicode strings. This is mostly a Python 3 problem, but could cause problems elsewhere.

To solve this problem, I've taken a similar approach to https://github.com/linsomniac/python-memcached/pull/100/files:

- Only byte strings get serialized without a flag
- Unicode strings get serialized with a different text flag
- Only specific builtin types get flags added, everything else gets pickled

Also `test_serde.py` wasn't being picked up with pytest as it didn't have a decorator - so I've added that.

### DIFF
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -23,6 +23,7 @@ except NameError:
     long_type = None
 
 
+FLAG_BYTES = 0
 FLAG_PICKLE = 1 << 0
 FLAG_INTEGER = 1 << 1
 FLAG_LONG = 1 << 2
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -36,18 +36,18 @@ def python_memcache_serializer(key, value):
 
     # Check against exact types so that subclasses of native types will be
     # restored as their native type
-    if value_type == six.binary_type:
+    if value_type is six.binary_type:
         pass
 
-    elif value_type == six.text_type:
+    elif value_type is six.text_type:
         flags |= FLAG_TEXT
         value = value.encode('utf8')
 
-    elif value_type == int:
+    elif value_type is int:
         flags |= FLAG_INTEGER
         value = "%d" % value
 
-    elif six.PY2 and value_type == long_type:
+    elif six.PY2 and value_type is long_type:
         flags |= FLAG_LONG
         value = "%d" % value
 
--- a/pymemcache/serde.py
+++ b/pymemcache/serde.py
@@ -14,6 +14,7 @@
 
 import logging
 from io import BytesIO
+import six
 from six.moves import cPickle as pickle
 
 try:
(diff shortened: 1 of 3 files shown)


pymemcache/serde.py
```diff
@@ -25,19 +26,31 @@
...
 
 
 def python_memcache_serializer(key, value):
     flags = 0
+    value_type = type(value)
 
-    if isinstance(value, str):
+    # Check against exact types so that subclasses of native types will be
+    # restored as their native type
+    if value_type == six.binary_type:
```
methane: (nits) I think `value_type is bytes` is more readable.

pymemcache/test/test_integration.py
```diff
@@ -232,6 +236,25 @@ def _des(key, value, flags):
...
+        result = client.get(b'key')
+        assert result == value
+
+    client = client_class((host, port), serializer=python_memcache_serializer,
+                          deserializer=python_memcache_deserializer,
+                          socket_module=socket_module)
+    client.flush_all()
+
+    check(b'byte string')
+    check(u'unicode string')
```
nichochar: Can we also check try some harder cases:
- `'olé'` (with and without specifying the `u` unicode prefix
- list of strings with unicode in them

pymemcache/test/test_integration.py
```diff
@@ -232,6 +236,25 @@ def _des(key, value, flags):
...
+    check(1)
+    check(123123123123123123123)
+    check({'a': 'pickle'})
+
+
+@pytest.mark.integration()
```
nichochar: 👍

pymemcache/test/test_serde.py
```diff
@@ -1,24 +1,58 @@
...
+        if not isinstance(serialized, six.binary_type):
+            serialized = six.text_type(serialized).encode('ascii')
+
         deserialized = python_memcache_deserializer(b'key', serialized, flags)
         assert deserialized == value
 
-    def test_str(self):
-        self.check('value')
+    def test_bytes(self):
+        self.check(b'value')
```
nichochar: let's try this with unicode characters.
Also maybe explicitely ask for `FLAG_BYTES` as it should default to this?

pymemcache/test/test_integration.py
tomkins: I've added in defaultdict (a subclass of dict) to test pickle comparisons as well - as it's available in Python 2.5+.

tomkins:
Not too sure what to do about the pypy3 failure.

The Travis build environment has changed between 15th-16th November. A newer version of virtualenv is installed, which is trying to install a newer version of pip - which now only supports Python 3.3+ for Python 3. The Travis environment is calling pypy3 v2.4.0 - a Python 3.2 implementation of Python.

I could just disable it, unless there's another option you prefer?

nichochar:
Thanks for raising this, I have opened [a ticket](https://github.com/pinterest/pymemcache/issues/132) in order to drop python 3.2.
I am leaning towards this option, but I would like to get a couple other opinions about it, we should have an answer by end of week.

nichochar:
After exploring this issue, we have decided to drop support for that interpreter version. Feel free to rebase and push, and we can go ahead with this PR.

tomkins:
Rebased and pushed.

nichochar:
Thanks for the improvements @tomkins tomkins, we really appreciate contributions!

PR #134, opened 2016-12-08
remove suggestion to use python-clandestined

ewdurbin:
this note is misleading, as the hasher is expected to expose a `get_node`
method. clandestined's `RendezvousHash` does not implement this method.

when clandestined's implementation was copied into pymemcached in commit
cff5abfad59baa593996fe7e240fd3944e2c4af2, the method was renamed from
`find_node` to `get_node`, breaking compatibility.


nichochar:
Awesome thanks for the cleanup!

PR #138, opened 2017-02-13
Raising MemcacheIllegalInputError when key contains null byte, newline, or carriage return

bwalks:
Memcache does not support setting keys with null byte, newline, carriage return. This adds support for checking the key to ensure the key is valid.

Setting keys with these returns the following exception:

```python
>>> client.set('\r\ntest', 'foo')
  File "lib/memcache.py", line 139, in set
    return super(Client, self).set(key, value, expire=expire, noreply=noreply)
  File .../env/lib/python2.7/site-packages/pymemcache/client/base.py", line 256, in set
    return self._store_cmd(b'set', key, expire, noreply, value)
  File .../env/lib/python2.7/site-packages/pymemcache/client/base.py", line 747, in _store_cmd
    self._raise_errors(line, name)
  File ".../env/lib/python2.7/site-packages/pymemcache/client/base.py", line 654, in _raise_errors
    raise MemcacheUnknownCommandError(name)
pymemcache.exceptions.MemcacheUnknownCommandError: set
```

This has been tested using version 1.3.5 on python 2.7.10

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -92,20 +92,27 @@ def _check_key(key, allow_unicode_keys, key_prefix=b''):
         except (UnicodeEncodeError, UnicodeDecodeError):
             raise MemcacheIllegalInputError("Non-ASCII key: '%r'" % (key,))
     key = key_prefix + key
-    if b' ' in key or b'\n' in key:
-        raise MemcacheIllegalInputError(
-          "Key contains space and/or newline: '%r'" % (key,)
-        )
-    if b'\00' in key:
-        raise MemcacheIllegalInputError(
-          "Key contains null byte: '%r'" % (key,)
-        )
-    if b'\r\n' in key:
-        raise MemcacheIllegalInputError(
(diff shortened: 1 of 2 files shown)


cgordon:
With this change, we are walking the entire key string three times to do validation. It's probably worth doing one walk, and checking each character along the way, rather than using "in" to do the work.

bwalks:
@cgordon I agree. I updated the commit to iterate over the characters once. I also split out the check for newline to enable a more specific error message.

bwalks:
The tests were failing because the character checks differ between python 2 and 3. Iterating over the binary string in python 2 returns characters, while python 3 returns the ordinal value.

nichochar:
This looks good to me, thanks for the catch, and the improved optimized traversing. The build is failing for an unrelated issue, which will be fixed in https://github.com/pinterest/pymemcache/pull/140

Once that's merged, please rebase and push, and we can go ahead and merge this.

bwalks:
@nichochar thanks for the update! Pulled in the latest changes.

PR #145, opened 2017-03-14
Return default value instead of False when all clients are down

renstrom:
#The documentation for `ignore_exc` says it will treat memcache/network errors as cache misses (i.e. `None` for reads and `False` for writes).

However the short-circuit path for when no client can be used (because they're all marked as down/bad) always returns `False`. This fix, changes the behaviour to return each command's default value instead.

I actually stumbled upon this because I wanted to inspect some behaviour in a system without cache so I stopped all memcached nodes. Our code looked something like this:

```python
data = memcached_client.get(key)
if data is None:
    return fetch_from_database(key)
else:
    return json.loads(data)
```

What happened was that every once in a while I would see a stacktrace because the `json.loads(data)` call would fail with a `TypeError` because it tries to deserialize a bool (`False`).

### DIFF
--- a/pymemcache/test/test_client_hash.py
+++ b/pymemcache/test/test_client_hash.py
@@ -177,7 +177,7 @@ class TestHashClient(ClientTestMixin, unittest.TestCase):
         with pytest.raises(socket.error):
             client.get('foo')
 
-    def test_no_servers_left_with_commands(self):
+    def test_no_servers_left_with_commands_return_default_value(self):
         from pymemcache.client.hash import HashClient
         client = HashClient(
             [], use_pooling=True,
@@ -186,6 +186,8 @@ class TestHashClient(ClientTestMixin, unittest.TestCase):
         )
 
         result = client.get('foo')
+        assert result is None
+        result = client.set('foo', 'bar')
         assert result is False
 
     def test_no_servers_left_with_set_many(self):
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -217,7 +217,7 @@ class HashClient(object):
         client = self._get_client(key)
 
         if client is None:
-            return False
+            return default_val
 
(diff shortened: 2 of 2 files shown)


pymemcache/test/test_client_hash.py
```diff
@@ -186,6 +186,8 @@ def test_no_servers_left_with_commands(self):
         )
 
         result = client.get('foo')
+        assert result is None
+        result = client.set('foo', 'bar')
         assert result is False
```
jogo: what about making sure get_many is `{}`?

renstrom: I assume you mean `client.get_many(['foo']) == {'foo': None}`? Currently it returns `{'foo': False}` (just tested). The former would make it more in line with how `.get(key)` works.

```python
result = client.get_many(['foo'])
assert result == {'foo': None}
```

jogo: I was referring to.
```>>> from pymemcache.client.hash import Client
>>> client = Client([],ignore_exc=True)
>>> client.get('foo')
>>> print client.get('foo')
None
>>> print client.get_many(['foo'])
{}
```

I wasn't able to reproduce your ```client.get_many(['foo']) == {'foo': None}```

renstrom: Well I just added a test at the end of the one I changed.

```diff
diff --git a/pymemcache/test/test_client_hash.py b/pymemcache/test/test_client_hash.py
index 30b2130..1b2709f 100644
--- a/pymemcache/test/test_client_hash.py
+++ b/pymemcache/test/test_client_hash.py
@@ -187,6 +187,8 @@ class TestHashClient(ClientTestMixin, unittest.TestCase):

         result = client.get('foo')
         assert result is False
+        result = client.get_many(['foo'])
+        assert result == {}
```

The above will fail.

Issue #103, opened 2016-07-19
Client fails for command: stats cachedump <slab> <itemcount>


rthille:
This line: https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L664
causes the 'stats cachedump 0 0' to turn into 'stats cachedump 0' or worse 'stats 0 cachedump' depending on the dict ordering.
Further, if you hack around that, the client doesn't know how to deal with the lines returned which start with 'ITEM '

nichochar:
Switching out the keys dict for a `collections.OrderedDict` should fix this. 
Thanks for reporting @rthille . Would you like to make a PR fix?

rthille:
I don't think just moving to an ordered dict will work because for the 'stats cachedump' call you can have two args which are identical.

jogo:
Confirmed this is still a reproducible issue.

PR #146, opened 2017-03-24
Add support for stats cachedump

jogo:
Stats cachedump has a different input and output format compared to
most stats commands. Previously this would fail with a cryptic error.

Fixes issue #103

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -705,8 +705,14 @@ class Client(object):
             raise MemcacheServerError(error)
 
     def _fetch_cmd(self, name, keys, expect_cas):
-        checked_keys = dict((self.check_key(k), k) for k in keys)
-        cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
+        if name == b'stats':
+            # stats commands can have multiple arguments
+            #   `stats cachedump 1 1`
+            checked_keys = [self.check_key(k) for k in keys]
+            cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
+        else:
+            checked_keys = dict((self.check_key(k), k) for k in keys)
+            cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
 
         try:
             if not self.sock:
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -705,8 +705,12 @@ def _raise_errors(self, line, name):
             raise MemcacheServerError(error)
 
     def _fetch_cmd(self, name, keys, expect_cas):
-        checked_keys = dict((self.check_key(k), k) for k in keys)
-        cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
+        if name == b'stats':
```
nichochar: Can we add a comment here explaining why the stats command warrants space splitting, for future readers.

pymemcache/client/base.py
```diff
@@ -744,6 +748,9 @@ def _fetch_cmd(self, name, keys, expect_cas):
                 elif name == b'stats' and line.startswith(b'STAT'):
                     key_value = line.split()
                     result[key_value[1]] = key_value[2]
+                elif name == b'stats' and line.startswith(b'ITEM'):
```
nichochar: Same here, can we add a quick comment explaining why ?

pymemcache/client/base.py
jogo: Done

pymemcache/client/base.py
jogo: Done

pymemcache/client/base.py
```diff
@@ -705,8 +705,14 @@ def _raise_errors(self, line, name):
             raise MemcacheServerError(error)
 
     def _fetch_cmd(self, name, keys, expect_cas):
-        checked_keys = dict((self.check_key(k), k) for k in keys)
-        cmd = name + b' ' + b' '.join(checked_keys) + b'\r\n'
+        if name == b'stats':
+            # stats commands can have multiple arguments
+            #   `stats cachedump 1 1`
+            checked_keys = [self.check_key(k) for k in keys]
```
jparise: This could also be `map(self.check_key, keys)`.

pymemcache/client/base.py
```diff
@@ -744,6 +750,10 @@ def _fetch_cmd(self, name, keys, expect_cas):
                 elif name == b'stats' and line.startswith(b'STAT'):
                     key_value = line.split()
                     result[key_value[1]] = key_value[2]
+                elif name == b'stats' and line.startswith(b'ITEM'):
+                    # For 'stats cachedump' commands
+                    key_value = line.split()
```
jparise: I think you could also use `line.split(b' ', 1)` here so you don't need to re-join the remainder below.

PR #149, opened 2017-04-19
Use StringIO for StringBuffer building

simonzheng:
Rather than allocating new strings with every string concatenation using normal string concatenation in base.py:Client. _store_cmd(), use StringIO for string building

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -786,7 +786,8 @@ class Client(object):
         if noreply:
             extra += b' noreply'
 
-        cmd_buffer = StringIO(name)
+        cmd_buffer = StringIO()
+        cmd_buffer.write(name)
         cmd_buffer.write(b' ')
         cmd_buffer.write(key)
         cmd_buffer.write(b' ')
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -14,6 +14,10 @@
 import errno
 import socket
 import six
+try:
+    from cStringIO import StringIO
+except ImportError:
+    import StringIO
 
 from pymemcache import pool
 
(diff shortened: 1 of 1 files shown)


simonzheng:
@cgordon @sontek @nichochar -- not sure who the maintainers are but adding you guys since you're the most active contributors here: https://github.com/pinterest/pymemcache/graphs/contributors

jparise:
@simonzheng: Many test failures like this:

```python
cStringIO.StringI' object has no attribute 'write'
```

simonzheng:
@jparise -- ah yeah, that was a very last second change that I made. fixed it and hopefully tests pass now :)

jogo:
Looks like another valid error

cgordon:
I'm not sure I understand the goal of this change. Is StringIO supposed to be faster that string concatenation? I see some evidence to indicate the opposite:

https://gist.github.com/djfroofy/3258823

If the goal of this change is to make the code more readable then I guess opinions differ, but this looks much less readable to me.

nichochar:
Hi, first of all, thanks for the PR @simonzheng ! We really appreciate them.

A few thoughts:
* is this part of the code a bottleneck? it's my opinion that creating a string here is negligible compared to communicating with the memcached service.
* still, if this is a speed improvement, I think we should merge it in
* I agree with @cgordon that we have lost readability, but I think we can fix that by simply putting the strings into a list and then calling like the following? I wonder if it's slower though...
```python
cmd_list =  [name, b' ', key,  b' ', six.text_type(flags).encode('ascii'), ...]
cmd_buffer = StringIO()
for elt in cmd_list: 
    cmd_buffer.write(elt)
```

Would it be possible to run benchmarks of your branch's solution vs the current branch solution, proving that this will indeed lead to a speed increase?

Thanks!

simonzheng:
Hi guys,

Thank you for the feedback and suggestions! I don't think this part of the code is the main bottleneck at all, but figured that any performance improvements that are low-effort merited a pull request submission :) The string concatenation was just something I noticed as I was combing through the pymemcache source code, though I would need to do performance benchmarks to confirm, especially after reading cgordon's attached link.

As for performance benchmarks, I won't be able to run them any time soon given that I'm working on a time-sensitive deadline but will return to this when I have a little more bandwidth personally. 

Thank you for taking the time to look at this and giving feedback 👍

jparise:
@simonzheng, are you up for revisiting this?

simonzheng:
Hey Jon,

At the moment I won't be able to due to vacation and impending deadlines
thereafter. Let's close this, since it was originally an optimization that
might not even work (requires more research). Thank you for following up.

Best,
Simon

PR #150, opened 2017-04-23
Remove docs from being run when you're running tests

nichochar:
Whenever you run tox locally, it builds the docs. This has the effect of building a bunch of new files that need cleaning.

Readthedocs builds the docs by itself, it doesn't need them to be checked in anyway, so this makes sure tox.ini is idempotent and doesn't modify any local files.


jogo:
Should we update `.gitignore` to work for folks who want to improve the docs?  So when `tox -edocs` is ran you don't have to clean files up etc.

nichochar:
I think we can, yes.
I believe all we need from the `docs` folder is `conf.py`, and we can keep the files that are manual:
* `getting_started.rst`
* `index.rst`

PR #158, opened 2017-06-02
Reverse backwards compatible change

nichochar:
We made a change that was backwards incompatible and didn't realize until after the merge. ALthough we are still interested in upgrading our pickle method, we need to think it through further in order to preserve backwards compatibilty


jogo:
Looks good

PR #160, opened 2017-06-03
Make serde pickle version configurable

akatrevorjay:
See https://github.com/pinterest/pymemcache/pull/156 for more info :)


akatrevorjay:
My quick hack was reverted. This provides similar functionality while maintaining backwards compatibility with ancient pythons or a mix of py2/py3 and python-memcached.

pymemcache/serde.py
```diff
@@ -17,82 +17,128 @@
...
+
+        elif flags & FLAG_PICKLE:
+            try:
+                buf = BytesIO(value)
+                unpickler = pickle.Unpickler(buf)
+                return unpickler.load()
+            except Exception as exc:
+                # This includes exc as a string for troubleshooting as well as providing
+                # a trace.
+                log.exception('Could not depickle value: %s')
```
akatrevorjay: I changed this because using the root logger is nearly impossible to troubleshoot, and this should really be exc level.

akatrevorjay:
fin

akatrevorjay:
How's this look everyone? Any thoughts?

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+          it in your pickleable state via `__getstate__`, python will raise:
+            ```
+            TypeError: a class that defines __slots__ without defining
+            __getstate__ cannot be pickled
+            ```
+        """
+        if pickle_version is not None:
+            self.pickle_version = pickle_version
+
+    def from_python(self, key, value):
```
nichochar: how about call this function `serialize()` (here and in metaclass)

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+            flags |= self.FLAG_PICKLE
+
+            output = BytesIO()
+            pickler = pickle.Pickler(output, self.pickle_version)
+            pickler.dump(value)
+            value = output.getvalue()
+
+        return value, flags
+
+    def to_python(self, key, value, flags):
```
nichochar: Same here with `deserialize`

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        Deserialize a value into a python object.
```
nichochar: Thanks for the doc

pymemcache/serde.py
```diff
@@ -1,98 +1,10 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+"""
+Backwards compatibility with the older serialization api previously
+provided by this module.
+"""
+from . import codecs
```
jogo: I prefer to use absolute imports here.

https://www.python.org/dev/peps/pep-0008/#imports

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2012 Pinterest.com
```
jogo: why rename the file?

pymemcache/serde.py
```diff
@@ -1,98 +1,10 @@
...
-from six.moves import cPickle as pickle
+_SERDE = codecs.Serde()
 
-try:
-    long_type = long  # noqa
-except NameError:
-    long_type = None
-
-
-FLAG_BYTES = 0
```
jogo: Although I don't think removing these from this file will be a problem it is technically a backwards incompatible change and don't want to break backwards compatibility.

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+    Serialization handler.
+
+    Meant to be compatible with `python-memcached`.
+    """
+
+    FLAG_BYTES = 0
+    FLAG_PICKLE = 1 << 0
+    FLAG_INTEGER = 1 << 1
+    FLAG_LONG = 1 << 2
+    FLAG_COMPRESSED = 1 << 3  # unused, to main compatibility with python-memcached
```
jogo:
```
./pymemcache/codecs.py:71:81: E501 line too long (83 > 80 characters)
    FLAG_COMPRESSED = 1 << 3  # unused, to main compatibility with python-memcached
```

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+
+        elif flags & self.FLAG_PICKLE:
+            try:
+                buf = BytesIO(value)
+                unpickler = pickle.Unpickler(buf)
+                return unpickler.load()
+            except Exception as exc:
+                # This includes exc as a string for troubleshooting as well as
+                # providing a trace.
+                log.exception('Could not depickle value (len=%d): %s',
```
jogo: unclear why this is better:

``` python
import logging
try:
    raise Exception("raised")
except Exception as exc:
    logging.exception("TEST: %s", exc)


try:
    raise Exception("raised")
except Exception:
    logging.exception("TEST")
```

```
ERROR:root:TEST: raised
  File "test.py", line 3, in <module>
    raise Exception("raised")
Exception: raised
ERROR:root:TEST
  File "test.py", line 9, in <module>
    raise Exception("raised")
Exception: raised
```

pymemcache/test/test_codecs.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+from unittest import TestCase
+
+from pymemcache import codecs
+import pytest
+import six
+
+try:
```
jogo: We should import this the same way we do elsewhere in this code: `from six.moves import cPickle as pickle`

pymemcache/test/test_codecs.py
```diff
@@ -0,0 +1,75 @@
...
+        self.check({'a': 'dict'}, self.Serde.FLAG_PICKLE)
+
+    def test_subtype(self):
+        # Subclass of a native type will be restored as the same type
+        self.check(CustomInt(123123), self.Serde.FLAG_PICKLE)
+
+    def test_pickle_version(self):
+        for pickle_version in range(-1, pickle.HIGHEST_PROTOCOL):
+            self.check(
+                dict(whoa='nelly', humans=u'horrid', answer=42),
```
jogo: I would prefer to use a different key value pair for humans.

pymemcache/codecs.py
```diff
@@ -0,0 +1,170 @@
...
+    FLAG_COMPRESSED = 1 << 3  # unused, to main compatibility with python-memcached
+    FLAG_TEXT = 1 << 4
+
+    pickle_version = 0
+
+    def __init__(self, pickle_version=0):
+        """
+        Init
+
+        :param int pickle_version: Pickle version to use (from python).
```
jogo: It's not the pickle version but rather the pickle protocol version

https://docs.python.org/3/library/pickle.html#data-stream-format

pymemcache/codecs.py
akatrevorjay: To provide backwards compatibility

akatrevorjay: (Without polluting the primary namespace)

akatrevorjay: It's also a better name.

pymemcache/codecs.py
akatrevorjay: Because from the *initial* line, you know what the exc was that caused it. This helps when an exception storm hits and you have N of them in your log interspersed.

akatrevorjay: Logging entries are usually configured with a better format than the default that you used, which contain things such as the process and thread ID.

pymemcache/test/test_codecs.py
akatrevorjay: While I can do that, it breaks tab completion.

pymemcache/test/test_codecs.py
akatrevorjay: Please assist me with what you would like.

pymemcache/codecs.py
akatrevorjay: I'll conform to VT100 in an update

pymemcache/codecs.py
akatrevorjay: Thanks, will update.

pymemcache/codecs.py
akatrevorjay: I changed it because in my experience serialize and deserialize are not the best names. Using a common anchor it seems to stick with more juniors easier and avoids headache. Still, I can update if you want, but I'd recommend not.

pymemcache/codecs.py
akatrevorjay: Nicochar, anytime for you bud.

pymemcache/serde.py
akatrevorjay: PEP8 allows relative imports. Is there a reason to not use them and opt for the less featureful alternative?

pymemcache/serde.py
akatrevorjay: IMO These should have been private to begin with, plus can't you just bump the major? I'll import them into serde.py if you want though.

pymemcache/test/test_codecs.py
akatrevorjay: Oh, you mean humans aren't horrid? But we are! Lol, I'll update it!

pymemcache/codecs.py
nichochar: thanks!

pymemcache/codecs.py
nichochar: I think this is fine.

nichochar:
Hi @akatrevorjay ,

After internal discussion:
* we really appreciate the contribution, and believe that making the cpickle version configurable is a good (great even) change.
* we want the change to make it to master, but we would like it to be as atomic and small as possible before we perform the merge.

Another way of saying this is that this is a rather low level library that many people depend on, and few people actually open up and modify. This means that the benefits of a refactor are rather low, whereas the risks of something breaking because of it would impact a large number of actors in a bad way.

Such a refactor would require us to probably bump a major version for the library, which doesn't make sense given the feature set change.

All of this being said, we all agree that your code is good and the new interface and naming make sense.

## Course of action:

We would like to open a new PR with as little changes as possible allowing for the configurable CPickle version. We can then bump that as a minor patch.

Since you've already done a lot of the work here, we're happy to make the PR changes on our side, but if you would like to do them on yours, let us know.

Cheers!

pymemcache/serde.py
akatrevorjay: done, consistency > opinions

pymemcache/codecs.py
akatrevorjay: done

pymemcache/codecs.py
akatrevorjay: done

pymemcache/codecs.py
akatrevorjay: done

pymemcache/test/test_codecs.py
akatrevorjay: done

pymemcache/codecs.py
akatrevorjay: done

akatrevorjay:
@nichochar How's that? Addressed all concerns, I believe?

pymemcache/serde.py
akatrevorjay: This is done, I imported them into `serde.py`. Let me know if this is not adequate for reasons though!

akatrevorjay: I already imported the API methods, and I don't think anyone should really be using constants in an arbitrary location in a library.

akatrevorjay:
The failure is due to #178 , ie nothing to worry about

akatrevorjay:
Updated to latest master so it can pass tests (my other PR fixed that)

nichochar:
Hello again, thanks for bumping! We definitely appreciate the work. 

Since there is a lot of changes, and I'm no longer making big decisions related to this project, I will defer the review decision to @jogo .

In of itself, the code looks fine, the arguments against it I believe are that the project is very mature and stable, this isn't fixing any bugs but refactoring quite a bit, and losing blame history with a file rewrite. The benefits need to be measured well, and I'll defer to @jogo

jparise:
Closing this in favor of the approach implemented in #190.

PR #162, opened 2017-07-25
Add support for UNIX sockets

dkuryakin:


jogo:
Thank you for the pull request,

Adding support for unix sockets makes a lot of since. It would be nice to add some documentation and testing around this change. If you would like we can add the tests and documentation for you.

dkuryakin:
Not a problem. Will add docs/tests soon

nichochar:
Hi @dkuryakin , we're excited about this contribution!, do you know when you will have a bit of free time to tackle a bit of docs/tests?

WolverineFan:
This works pretty well, but you'll also want to avoid the `setsockopt` call when `no_delay` is set for a unix socket.

jparise:
Closing in favor of #206.

PR #164, opened 2017-09-11
Don't Raise Generic Exception

njharman:
Libraries should raise library specific exceptions. Such as
MemcacheError. This enables users to easily catch exceptions generated
by library separate from other exceptions. Raising "Exception" forces
catching Exception and potentially masks all sorts of issues.


jparise:
I think there's a line of unit test code that should be updated to use the more-specific `MemcacheError` here, too.

https://github.com/pinterest/pymemcache/blob/fe241ceafc317f40b7a03d514279397c3c7e731f/pymemcache/test/test_client_hash.py#L165

PR #165, opened 2017-10-06
Close client socket if it fails to connect

cjwatson:
If Client._connect fails, it should close the socket before re-raising
the exception; the caller can't do this because the socket is only
stored in a local variable until it's been fully connected.  Failing to
do this causes ResourceWarnings in Python 3.


cjwatson:
Thanks for the super-quick review!

PR #168, opened 2017-11-15
Update getting_started.rst

gs230937207:
With python3 (3.5 here) json_deserializer receives the values as bytes. The added decode calls in json_deserializer take care of that problem.

Disclaimer: I am a python beginner, so if there is another way to achieve the same result please let me know.


nichochar:
This is great, we really appreciate documentation help, thanks a lot @gs230937207

PR #170, opened 2017-11-19
Added Python 3.5 and 3.6 to the build

thedrow:


thedrow:
Seems like we need to rearrange the matrix :(

jogo:
@thedrow thanks for the PR, we definitly should be testing against python 3.5 and 3.6.

I agree we need to rearrange the test matrix, would you like to do that, if not I would be happy to help.

We should also update the python classifiers in setup.y to reflect that we support 3.5 and 36.

thedrow:
I added it to my TODO list. If I don't get to it in the next few days feel free to go ahead.

jogo:
@thedrow https://github.com/pinterest/pymemcache/pull/171

jogo:
merged as part of #171

PR #171, opened 2017-11-27
Update testing matrix

jogo:
* Test against python 3.5 and 3.6
* Run flake8 in travis
* Add missing python versions to travis config, previously some tests passed with:
    
        ERROR: InterpreterNotFound: python2.6
        SKIPPED:  py26: InterpreterNotFound: python2.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ matrix:
       env: TOXENV=py36,py36-flake8
     - python: pypy
       env: TOXENV=pypy
-# Drop pypy3 support until travis fixes python3 pypy interpreter path issue
-# https://github.com/travis-ci/travis-ci/issues/6304#ref-issue-177592739
+    - python: pypy3
+      env: TOXENV=pypy3
 
 services:
   - memcached
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,27 @@ cache:
   directories:
     - $HOME/.cache/pip
 language: python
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -3,22 +3,23 @@ cache:
...
+    - "2.6"
+    - "2.7"
+    - "3.3"
+    - "3.4"
+    - "3.5"
+    - "3.6"
+    - "pypy"
+    - "pypy3"
+# Drop support until travis fixes python3 pypy interpreter path issue
+# https://github.com/travis-ci/travis-ci/issues/6304#ref-issue-177592739
```
jparise: Are these comments still correct if `pypy3` is now in the list?

This seems more relevant to the tox configuration, which doesn't include `pypy3`.

jogo: Looks like the comment is outdated.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, pypy, py33, py34, py35, py27-flake8, py35-flake8, integration
+envlist = py27, pypy, py33, py34, py35, py27-flake8, py36-flake8, integration
```
jparise: `pypy3` doesn't need to appear here?

jogo: envlist is used when `TOXENV` isn't set, which only happens when testing locally.

It should be here to make the local test list similar to travis.

PR #178, opened 2018-06-19
Remove deprecated Python version. This makes tests fail at the moment.

akatrevorjay:
This is because `wheel` requires it *not* be `3.3.x`.

Of course, we can instead hack this to install some ancient wheel on an
ancient Python... but nobody should be running `3.3.x`, so it'd
just be cruft, at least in my garbage opinion ;)


nichochar:
I think this is a good move. @jogo anything opposed to me merging this?

PR #179, opened 2018-07-10
fix django2.0

opapy:
Failed `set_many` with django2.0.

```python
>>> from django.core.cache import cache
>>> cache.set_many({'aa': 1})
  File "/usr/lib/python3.6/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/ddtrace/contrib/django/cache.py", line 65, in wrapped
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django_elastipymemcache/memcached.py", line 43, in wrapper
    return f(self, *args, **kwds)
  File "/usr/local/lib/python3.6/dist-packages/django_elastipymemcache/memcached.py", line 135, in set_many
    return super(ElastiPyMemCache, self).set_many(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django/core/cache/backends/memcached.py", line 144, in set_many
    return [original_keys[k] for k in failed_keys]
TypeError: 'bool' object is not iterable
```

Release Note: https://docs.djangoproject.com/en/2.0/releases/2.0/#cache
Code: https://github.com/django/django/blob/338f741c5eb6b91118f6a6b7c34b5e9b47a5661d/django/core/cache/backends/memcached.py#L131-L139

Since django 2.0, the return value of set_many now expects a list of keys of failed objects in store.

It should be implemented at the cache backend, but python-memcached and pylibmc return list objects, so I tried it accordingly.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -310,6 +310,7 @@ class Client(object):
 
         Returns:
           Returns a list of keys that failed to be inserted.
+          If noreply is True, alwais returns empty list.
         """
         # TODO: make this more performant by sending all the values first, then
         # waiting for all the responses.
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -309,7 +309,7 @@ class Client(object):
                    self.default_noreply).
 
         Returns:
-            Empty list
+          Empty list
         """
 
         # TODO: make this more performant by sending all the values first, then
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -309,17 +309,14 @@ class Client(object):
                    self.default_noreply).
 
         Returns:
-          If no exception is raised, always returns True. Otherwise all, some
-          or none of the keys have been successfully set. If noreply is True
-          then a successful return does not guarantee that any keys were
-          successfully set (just that the keys were successfully sent).
+            Empty list
         """
 
         # TODO: make this more performant by sending all the values first, then
         # waiting for all the responses.
         for key, value in six.iteritems(values):
             self.set(key, value, expire, noreply)
-        return True
+        return []
 
     set_multi = set_many
 
@@ -930,7 +927,8 @@ class PooledClient(object):
 
     def set_many(self, values, expire=0, noreply=None):
         with self.client_pool.get_and_release(destroy_on_fail=True) as client:
-            return client.set_many(values, expire=expire, noreply=noreply)
+            client.set_many(values, expire=expire, noreply=noreply)
+            return []
 
     set_multi = set_many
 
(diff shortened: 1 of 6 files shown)


jogo:
Looks good, although this does change the behavior of the API so we should make sure to document that in the changelog. but the change behavior is fairly minor and looks reasonable. This doesn't seem to fully implement django's expected behavior of returning a list of failed keys, is that expected?

opapy:
hi @jogo 
Sure, this implementation returns a list of keys failed to store in HashClient only.

PooledClient and Client always return empty list.
Client and PooledClient is always True, if it fails it returns an exception, so I thought whether this implementation would be a problem.

pymemcache/client/hash.py
```diff
@@ -261,11 +261,11 @@ def set_many(self, values, *args, **kwargs):
             new_args.insert(0, values)
             result = self._safely_run_func(
                 client,
-                client.set_many, False, *new_args, **kwargs
+                client.set_many, values.keys(), *new_args, **kwargs
```
jogo: so this means, if any of these keys fail, we mark all as failed?

opapy: thanks! 
That's true. I will fix it.

pymemcache/client/base.py
```diff
@@ -29,7 +29,7 @@
 
 RECV_SIZE = 4096
 VALID_STORE_RESULTS = {
-    b'set':     (b'STORED',),
+    b'set':     (b'STORED', b'NOT_STORED'),
```
opapy: python-memcached allow 'NOT_STORED'

Surgo:
@jogo any update on this issue?
This PR is blocker of django 2.0 migrations...

pymemcache/client/base.py
```diff
@@ -309,17 +309,19 @@ def set_many(self, values, expire=0, noreply=None):
                    self.default_noreply).
 
         Returns:
-          If no exception is raised, always returns True. Otherwise all, some
-          or none of the keys have been successfully set. If noreply is True
-          then a successful return does not guarantee that any keys were
-          successfully set (just that the keys were successfully sent).
+          Empty list
```
jogo: doesn't this now return a list of failed sets.

opapy: thx! i fixed it.

opapy:
pls restart job
https://travis-ci.org/pinterest/pymemcache/jobs/416230799

nichochar:
Job has been restarted

jogo:
Thanks looks good. Will work on releasing this sometime next week as a major version bump.

opapy:
@jogo
thunks for review!

Surgo:
@jogo When release it?

jparise:
@Surgo I think we're on track for a 1.5.0 release pretty soon. I'd like to see #184 reviewed and merged first, and then we can start preparing a new release.

PR #186, opened 2018-08-31
Add official Python 3.7 support

jparise:
Travis's default base image doesn't support Python 3.7 very well so we
need to explicitly select the xenial distribution with sudo enabled.


jogo:
looks good

PR #190, opened 2018-09-07
Make pickle version for python_memcache_serializer adjustable

jogo:
It's unsafe to use the max pickle version when you are switching between
versions of python with different max versions.

Add a new function get_python_memcache_serializer_pickle_version that
returns a python_memcache_serializer with any pickle version.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -43,17 +43,45 @@ Serialization
              return value, 1
          return json.dumps(value), 2
 
-    def json_deserializer(key, value, flags):
+     def json_deserializer(key, value, flags):
         if flags == 1:
             return value
         if flags == 2:
             return json.loads(value)
         raise Exception("Unknown serialization format")
 
-    client = Client(('localhost', 11211), serializer=json_serializer,
-                    deserializer=json_deserializer)
-    client.set('key', {'a':'b', 'c':'d'})
-    result = client.get('key')
+     client = Client(('localhost', 11211), serializer=json_serializer,
+                     deserializer=json_deserializer)
+     client.set('key', {'a':'b', 'c':'d'})
+     result = client.get('key')
+
+pymemcache provides a default
+`pickle <https://docs.python.org/3/library/pickle.html>`_-based serializer:
+
+.. code-block:: python
+
+    from pymemcache.client.base import Client
+    from pymemcache import serde
+
+    class Foo(object):
+      pass
+
+    client = Client(('localhost', 11211),
+        serializer=serde.python_memcache_serializer,
+        deserializer=serde.python_memcache_deserializer)
+    client.set('key', Foo())
+    result client.get('key')
+
+The serializer uses the highest pickle protocol available. In order to make
+sure multiple versions of Python can read the protocol version, you can specify
+the version with :code:`get_python_memcache_serializer`
+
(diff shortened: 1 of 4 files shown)


pymemcache/test/test_serde.py
```diff
@@ -59,3 +61,18 @@ def test_pickleable(self):
     def test_subtype(self):
         # Subclass of a native type will be restored as the same type
         self.check(CustomInt(123123), FLAG_PICKLE)
+
+
```
jparise: Worth adding a distinct test case for `pickle.HIGHEST_PROTOCOL`, too? It's redundant with `2`, but not in the "test the full interface" sense.

pymemcache/serde.py
```diff
@@ -63,13 +64,22 @@ def python_memcache_serializer(key, value):
...
         output = BytesIO()
-        pickler = pickle.Pickler(output, PICKLE_VERSION)
+        pickler = pickle.Pickler(output, pickle_version)
         pickler.dump(value)
         value = output.getvalue()
 
     return value, flags
 
 
+def get_python_memcache_serializer_pickle_version(pickle_version=None):
```
jparise: I think this could be shortened to `get_python_memcache_serializer()`, in case there are other reasonable parameters to provide in the future. That will keep the public API more stable.

pymemcache/serde.py
```diff
@@ -63,13 +64,22 @@ def python_memcache_serializer(key, value):
...
+    """Return a serializer using a specific pickle version"""
+    return partial(_python_memcache_serializer, pickle_version=pickle_version)
+
+
+python_memcache_serializer = partial(
+    _python_memcache_serializer, pickle_version=DEFAULT_PICKLE_VERSION)
```
jparise: This could just call `get_python_memcache_serializer`.

pymemcache/serde.py
jogo: good idea, I am bad at naming things.

jogo:
added a few lines to getting_started to cover the serde module

docs/getting_started.rst
```diff
@@ -43,17 +43,44 @@ Serialization
...
-    client = Client(('localhost', 11211), serializer=json_serializer,
-                    deserializer=json_deserializer)
-    client.set('key', {'a':'b', 'c':'d'})
-    result = client.get('key')
+     client = Client(('localhost', 11211), serializer=json_serializer,
+                     deserializer=json_deserializer)
+     client.set('key', {'a':'b', 'c':'d'})
+     result = client.get('key')
+
+pymemcache provides a default pickle based serializer:
```
jparise: I would link `pickle` to the Python docs here, and also hyphenate "pickle-based".

docs/getting_started.rst
```diff
@@ -43,17 +43,44 @@ Serialization
...
+      pass
+
+    client = Client(('localhost', 11211),
+        serializer=serde.python_memcache_serializer,
+        deserializer=serde.python_memcache_deserializer)
+    client.set('key', Foo())
+    result client.get('key')
+
+The serializer uses the highest pickle protocol available. In order to make
+sure multiple versions of python can read the protocol version you can specify
```
jparise: Comma after "protocol version" and capitalize Python.

PR #196, opened 2018-10-29
Added unix socket support

Morreski:
Hi,

Saw that PR #162 was pending (since 2017!)because tests and doc wasn't updated. Since I need this feature too, I made a slightly different version with test and docstring updated.

Sorry if the duplicate feature does not meet your contribution standards

Regards,

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -194,7 +194,7 @@ class Client(object):
         Constructor.
 
         Args:
-          server: tuple(hostname, port)
+          server: tuple(hostname, port) or string corresponding to a UNIX socket path.
           serializer: optional function, see notes in the class docs.
           deserializer: optional function, see notes in the class docs.
           connect_timeout: optional float, seconds to wait for a connection to
@@ -237,6 +237,7 @@ class Client(object):
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
         self.allow_unicode_keys = allow_unicode_keys
+        self._use_unix_socket = isinstance(server, str)
 
     def check_key(self, key):
         """Checks key and add key_prefix."""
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -237,15 +237,19 @@ def __init__(self,
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
         self.allow_unicode_keys = allow_unicode_keys
+        self._use_unix_socket = isinstance(server, str)
```
jparise: Perhaps we could set `self.socket_protocol` here in the initializer instead, and then use that directly in `_connect()` below. The `self._use_unix_socket` flag isn't otherwise used.

Alternatively, we could inspect the type of `self.server` in `_connect()` and only set `protocol` locally within that function.

pymemcache/test/test_client.py
```diff
@@ -71,6 +72,24 @@ def setsockopt(self, level, option, value):
         self.socket_options.append((level, option, value))
 
 
+class MockUnixSocket(object):
+
+    def __init__(self, socket_path):
+        if os.path.exists(socket_path):
+            os.remove(socket_path)  # Make sure no stalling connection exists
```
jparise: "stalled connection" might be clearer wording?

jparise:
@Morreski just wanted to check in and see if my feedback makes sense.

pymemcache/test/test_client.py
Morreski: Hi, sorry for the ghosting. Been pretty busy on other projects. All your statements make sense and sorry for the auto-formatting (I forgot to disable it).
I'll try to find some time to apply the modifications you suggested.

Cheers,

jparise:
Closing in favor of #206.

PR #197, opened 2018-10-31
Allow classes, functions, exceptions to be conveniently imported from the package.

mainanick:
The PR allows package level import
```python
from pymemcache import Client, PooledClient, MemcacheError, ...
```

Currently its
```python

from pymemcache.client.base import Client, PooledClient
or 
from pymemcache.client import Client, MemcacheError
```


mainanick:
Tests seems to fail with a "weird" error
 `Command "python setup.py egg_info" failed with error code 1. ImportError: No module named six`

PR #199, opened 2018-11-02
Fixes tox six pkg ImportError

mainanick:
This fixes travis ci invocation error caused by tox not installing some deps

### DIFF
--- a/pymemcache/__init__.py
+++ b/pymemcache/__init__.py
@@ -3,3 +3,11 @@ __author__ = "Charles Gordon"
 
 from pymemcache.client import Client  # noqa
 from pymemcache.client import PooledClient  # noqa
+
+from pymemcache.exceptions import MemcacheError  # noqa
+from pymemcache.exceptions import MemcacheClientError  # noqa
+from pymemcache.exceptions import MemcacheUnknownCommandError  # noqa
+from pymemcache.exceptions import MemcacheIllegalInputError  # noqa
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -3,6 +3,9 @@ envlist = py27, py34, py35, py36, py37, pypy, pypy3, py27-flake8, py37-flake8, i
 skip_missing_interpreters = True
 
 [testenv]
+setenv =
+    PYTHONPATH = {toxinidir}
+deps = -r{toxinidir}/test-requirements.txt
 commands =
     pip install -r test-requirements.txt
```
jparise: I don't think we still need this command if we're specifying the dependencies above.

mainanick: Do you mean `commands = pip install -r test-requirements.txt`?

jparise: Yes. I believe using `deps` implies a `pip install` command.

pymemcache/__init__.py
```diff
@@ -1,2 +1,5 @@
 __version__ = '2.0.0'
 __author__ = "Charles Gordon"
+
+from pymemcache.client.base import Client  # noqa
+from pymemcache.client.base import PooledClient  # noqa
```
jparise: Let's commit these lines as part of their own PR (#197) and focus just on fixing the `six` error here. Can you remove these from this PR?

PR #172, opened 2017-11-29
Enabling memory management from Client

bsb20:
Additional parts of the memcached API needed to be implemented to allow manual memory management from the Client.  In most cases these changes are incredibly simple.  The only core change required was that command arguments in _fetch_cmd always be sent in the order in which they are passed by the caller (arguments were pulled from a dictionary in arbitrary order), which has been implemented.

PR #211, opened 2019-01-08
Add support for the 'cache_memlimit' command

jparise:
This command allows clients to set a new cache memory limit. The
argument is a number expressed in megabytes.

Also, treat `OK` as an acceptable response string.

This is partially based on #172.


YacobBY:
Hello, would it be possible to add this memlimit function to the PooledClient?

PR #221, opened 2019-03-01
Introduce interactive CLI to use pymemcache Client from shell.

4383:
Introduce interactive/non-interactive CLI to use pymemcache from CLI.

These changes allow user to get access to pymemcache.base.Client methods
directly from shell in an interactive/non-interactive way.

Examples of use:

```
$ pymemcache --host 0.0.0.0
Pymemcache CLI 2.2.0
Connected to 0.0.0.0:11211
Memcache server version: 1.5.12
> help
...
> add foo bar
True
> get foo
b'bar'
> add -h
Usage (add): key, value, expire=0, noreply=None
Description:

The memcached "add" command.

Args:
  key: str, see class docs for details.
  value: str, see class docs for details.
  expire: optional int, number of seconds until the item is expired
           self.default_noreply).

Returns:
  If noreply is True, the return value is always True. Otherwise the
  return value is True if the value was stored, and False if it was
  not (because the key already existed).
> exit
$ pymemcache --host 0.0.0.0 -c "add -h"
Pymemcache CLI 2.2.0
Connected to 0.0.0.0:11211
Memcache server version: 1.5.12
Usage (add): key, value, expire=0, noreply=None
Description:

The memcached "add" command.

Args:
  key: str, see class docs for details.
  value: str, see class docs for details.
  expire: optional int, number of seconds until the item is expired
           self.default_noreply).

Returns:
  If noreply is True, the return value is always True. Otherwise the
  return value is True if the value was stored, and False if it was
  not (because the key already existed).
$ pymemcache --host 0.0.0.0 -c "add test test-value" -q
True
$ pymemcache --host 0.0.0.0 -c "get test" -q
b'test-value'
```


4383:
Fully functional but I guess some things can by adjust so review and comments are welcomes. 

:)

jparise:
@4383, this looks neat! It might take me a couple of days to try it out and review it though.

4383:
Can be tested by:

```shell
git clone https://github.com/4383/pymemcache -b introduce-cli
cd pymemcache
python setup.py install
docker run --publish 11211:11211 --name my-memcache -d memcached
pymemcache --host <container-ip>
```

4383:
@jparise Thanks. No problem take your time and do not hesitate to ask questions if needed :)

4383:
Some example bellow.

General command options:
```shell
pymemcache --help
usage: pymemcache [-h] [--host HOST] [-p PORT] [-c CMD] [-q] [-v]

Pymemcache command line client. Set and get values from a memcache server or
cluster. Using pymemcache 2.2.0

optional arguments:
  -h, --help            show this help message and exit
  --host HOST           Memcache host address
  -p PORT, --port PORT  Memcache server port (default 11211)
  -c CMD, --cmd CMD     Memcache command to execute
  -q, --quiet           Only print result command output
  -v, --version         Display pymemcache version and quit
```

Get available commands and help:
```shell
pymemcache --host 0.0.0.0 -c "help"                                         
Pymemcache CLI 2.2.0
Connected to 0.0.0.0:11211
Memcache server version: 1.5.12
Available commands:
        - add: The memcached "add" command
        - append: The memcached "append" command
        - cache_memlimit: The memcached "cache_memlimit" command
        - cas: The memcached "cas" command
        - check_key: Checks key and add key_prefix
        - close: Close the connection to memcached, if it is open
        - decr: The memcached "decr" command
        - delete: The memcached "delete" command
        - delete_many: A convenience function to delete multiple keys
        - delete_multi: A convenience function to delete multiple keys
        - flush_all: The memcached "flush_all" command
        - get: The memcached "get" command, but only for one key, as a convenience
        - get_many: The memcached "get" command
        - get_multi: The memcached "get" command
        - gets: The memcached "gets" command for one key, as a convenience
        - gets_many: The memcached "gets" command
        - incr: The memcached "incr" command
        - prepend: The memcached "prepend" command
        - quit: The memcached "quit" command
        - replace: The memcached "replace" command
        - set: The memcached "set" command
        - set_many: A convenience function for setting multiple values
        - set_multi: A convenience function for setting multiple values
        - stats: The memcached "stats" command
        - touch: The memcached "touch" command
        - version: The memcached "version" command
        - h: Display this helping message
        - help: Display this helping message
        - bye: Stop this pymemcache client
        - stop: Stop this pymemcache client
Display help command for available commands by using:
        <command> -h
Example:
        add -h
```

Interactive way:
```shell
pymemcache --host 0.0.0.0                                                                                                             
Pymemcache CLI 2.2.0                                                                                                                                                                                                 
Connected to 0.0.0.0:11211                                                                                                                                                                                           
Memcache server version: 1.5.12                                                                                                                                                                                      
> help                                                                                                                                                                                                               
Available commands:                                                                                                                                                                                                  
        - add: The memcached "add" command                                                                                                                                                                           
        - append: The memcached "append" command
        - cache_memlimit: The memcached "cache_memlimit" command
        - cas: The memcached "cas" command
        - check_key: Checks key and add key_prefix
        - close: Close the connection to memcached, if it is open
        - decr: The memcached "decr" command
        - delete: The memcached "delete" command
        - delete_many: A convenience function to delete multiple keys
        - delete_multi: A convenience function to delete multiple keys
        - flush_all: The memcached "flush_all" command
        - get: The memcached "get" command, but only for one key, as a convenience
        - get_many: The memcached "get" command
        - get_multi: The memcached "get" command
        - gets: The memcached "gets" command for one key, as a convenience
        - gets_many: The memcached "gets" command
        - incr: The memcached "incr" command
        - prepend: The memcached "prepend" command
        - quit: The memcached "quit" command
        - replace: The memcached "replace" command
        - set: The memcached "set" command
        - set_many: A convenience function for setting multiple values
        - set_multi: A convenience function for setting multiple values
        - stats: The memcached "stats" command
        - touch: The memcached "touch" command
        - version: The memcached "version" command
        - h: Display this helping message
        - help: Display this helping message
        - bye: Stop this pymemcache client
        - stop: Stop this pymemcache client
Display help command for available commands by using:
        <command> -h
Example:
        add -h
> add test test-value
True
> get test
b'test-value'
> add -h
Usage (add): key, value, expire=0, noreply=None
Description:

The memcached "add" command.

Args:
  key: str, see class docs for details.
  value: str, see class docs for details.
  expire: optional int, number of seconds until the item is expired
           self.default_noreply).

Returns:
  If noreply is True, the return value is always True. Otherwise the
  return value is True if the value was stored, and False if it was
  not (because the key already existed).
> exit
$
```

4383:
I guess I'll add the command `-d/--decode` option to decode output command.

4383:
I need to fix some PEP8 errors

4383:
PEP8 fixes done

4383:
I also need to add some dedicated unit test but not today :)

4383:
Possible enhancement:
- introduce [mypy](https://github.com/python/mypy) to annotate methods signatures and more properly retrieve/transform args and theirs types.
- decode output
- add unit tests
- improvement documentation

Thoughts/Ideas?

4383:
Don't spend your time review it now I want to changes a lot of things.

4383:
Still in development (I'm a little bit busy since 2 weeks) but some tiny news:
- refactor to represent commands/interpreter as an object

I need to annotate `Client` to properly parse args, I guess I'll use argparse internally to generate UI of the interpreter, and also I need to use [`shlex`](https://docs.python.org/3/library/shlex.html) to parse user command line and use it with argparse under the hood.

Argparse will be init with `Client` method annotations.

I guess this is most  agnostic way to keep `Client` completely decoupled of UI and interpretor.

You just need to maintains and evolve on Client and interpreter is automatically sync with new mechanism without maintains this part (interpretor) daily.

The only few things that you have to do is to ensure that Client methods provide annotations.

Thoughts?

4383:
Well, start to sound good, I've implemented all the previous things described in the previous comment:
- user input parsing with shlex
- annotate Client
- using argparse with each Client methods signatures

Everything seem works.

Now I need to provide a better output command management when results are type of list, tuple etc... and allow users to use raw result format if needed.

The interactive process seem ok I need to double check the pure CLI process (only one command  exec and quit).

You can check my work by using:
```sh
git clone git@github.com:4383/pymemcache -b introduce-cli
cd pymemcache
python -m venv .
source bin/activate
python setup.py develop
# run a memcache server
pymemcache --host 0.0.0.0
```

jparise:
Off to a great start! I was able to drive around it a bit locally, and it works really well.

We still want to maintain Python 2 compatibility so we'll need to do without the Python 3 type annotations though.

4383:
@jparise thanks.

So to keep python2 compatibility I guess we need to use mypy instead of stdlib annotations style.
I need to dig into mypy but since I already provide `utils` module I guess I just need to introduce some little changes on it (utils module) and everything will works fine.

Also I've check the CLI mode (not interactive mode) and everything works fine.

I've introduce raw input but I need to finish some changes on it.

After that I need to improve doc, fix pep8 errors, continue to add unit tests.

FYI, using argparse and shlex allow us to send commands like this:
```
# Consider run the Client add  command
$ pymemcache -host 0.0.0.0
> add key --expire 60 value
True
> get key
value
> add key2 value2 --expire 60
True
> get key2
value2
> add key3 "long string with blank space blablabla" --expire 60
True
> get key3
long string with blank space blablabla
> exit
```

We can pass parameters in different order without problems like a standard CLI.

Also I've implemented the `inspect` command who allow us to debug/inspect available commands and retrieve a lot of useful informations about these commands (object name of the method, params, argparse, etc...).

4383:
Last push summary:
- fixing methods who accept list parameters (`get_multi`, `get_many`) due to argparse side effect
- introduce the `history` command
- introduce the possibility to run command from history by using index (like bash history behavior)

Examples:

```
>>> add test test1
>>> get test
test1
>>> history
1 add test test1
2 get test
>>> history
1 add test test1
2 get test
3 history
>>> !2
test1
>>> exit
```

The current command is added to the history after execution. If the command is not found or is not a valid command we ignore it in history.

History is only for the current session, if the interpreter is closed the history is not stored, we can eventually store it in the home of users but we need to implement this mechanism.

jparise:
Given the scope (and ambition) of this interface, I'm starting to wonder if it would be better off as its own companion project. Thoughts?

4383:
@jparise can make sense.

In the first time I've imagined to create a separted project but the original purpose of this feature is provide debug facilities for projects who use pymemcache in production environment this is why I prefer to join the lib instead of split it to an another project.

Thoughts?

4383:
Projects like openstack have specifc constraints on requirements etc... and production environments are often locked on the dependencies side so this is why I prefer see the interpreter embdded in the lib with a separated entry point.

4383:
Another reason to stay closer from the lib is the fact that the interpreter is entirely generated by using introspection on the `Client` methods (currently annoations, soon mypy) and I guess we need to take care of the future of this class and the evolutions on it if we want still to adopt them on the interpreter side.

IHMO I'm afraid if we put the interpreter outisde the project (the lib) we don't take care about "annotations" (meaning mypy too) and we risk to loose the couple. 

Thoughts?

4383:
After inspect the mypy alternative to keep backward compatiblity with python2 and after thinking about the pros/cons of the companion way I guess we can consider to wait few months before introduce officially these changes.

Also the most part of distros now use python3 and python 2 [will not be maintained in 9 months](https://pythonclock.org/) , I guess is useless to spend time to keep backward compatibility.

I propose to you to wait for python 2 abandon instead of using things who will be deprecated in few months. I guess this is the most logical approach.

About the companion, the problem is the same while python 2 is alive due to the fact that we need to add changes to annotate the `Client` side.

When python 2 will be dead we will can consider to continue on the lib or to start with the companion.

During this time I can continue to evolve the interpreter on my side by rebasing through the changes.

We  could plan a _python 3_ gap/step in few months where we will remove `six`,  merge the annotations, and quit completely the python 2 compatibility and decide to start or not start companion.

Dilemma summarize below... 
Pros:
- stdlib annotations is straightforward to use...
- no external dependencies needed changes only use stdlib and best practices
- python 2 support will be stop in few months
Cons:
- mypy doesn't seem obvious to use... 
- 2 external dependencies needed
- network needed to install pymemcache (dependencies installation) (clone install or [COTS install](https://en.wikipedia.org/wiki/Commercial_off-the-shelf)[1] vs pypi install)
- maybe other things that I forget here...

Make sense ? Thoughts ?

[1] Some customers fork projects, own, and maintains packages repositories (rpm, deb, etc..) (often to enforce security rules or due to an locked production environment or things like that) etc... as many dependencies we need, as many customers need to maintain COTS. I guess the KISS philosophy (Keep It Simple (and Stupid why not))  must prime. Remember the six issues :)

jparise:
Once we (Pinterest) have all of our environments running on Python 3, we can consider releasing a Python 3-only version of pymemcache as a new major version while maintaining the existing Python 2-compatible version as a separate branch. Barring other considerations, I don't expect we'll want to do that until next year though.

At the moment, it looks like this new feature is the only thing that would require us to drop Python 2 support, which unfortunately isn't very strong motivation on its own. 🙁

4383:
@jparise sure make sense, and we are on the same frequency :)
This feature can wait for the moment and continue to be designed/evolved on my own fork and merged in the future, no stress for you and me.

The python-2 branch is a good approach and we can properly inform users by releasing a major[1] (version (3.0.0) and keep (2.X.X) to backport for python 2 by releasing minor/patches[1] versions on it if we add new features and fixes.

I don't know if you have github label for this (I guess no) to help users to better identify the python 3  choice, but if you can I guess it's better to label it.

I prefer keep the pull request opened on WIP mode to rapidly identify conflicts and avoid them.

[1] according to [semver](https://semver.org/)

jparise:
@4383 now that we're in a new major version (3), any interest in resuming this work?

4383:
@jparise first thanks for the heads up. Yep it could be interesting to unearth these changes and to continue with this topic. I'll try to spend time on it during the next weeks.

4383:
Sorry for the black hole, I'll try to restart work on that

4383:
Hello,

The latest version of these changes is a complete refactor of my previous approach.

I decided to rely on tox's capabilities to allow us to emulate an interpreter.

These changes allow us to interact with pymemcache and memcached servers directly from our shell by using tox's virtualenv.
    
By following this way we don't need to maintain a specific CLI, which mean:
- a "CLI" always up-to-date and sync-up with the lib;
- less technical debt;
- no code to maintain;
- a command history directly managed by the python interpreter;
- less workload for the maintainers of pymemcache;
-  less chances to face bugs and side effects;
- a better access to the python debugger.

Less is more.

docs/getting_started.rst
```diff
@@ -174,6 +174,79 @@ the `JsonSerde` from above which is more careful with encodings:
...
+          key: str, see class docs for details.
+          default: value that will be returned if the key was not found.
+        Returns:
+          The value for the key, or default if the key wasn't found.
+
+You can instantiate all the class and clients offered by pymemcache.
+
+Your client will remain open until you decide to close it or until you decide
+to quit your interpreter. It can allow to see what's happen if your server
+is abruptely closed. Below is an by example.
```
jparise, line 204:
```suggestion
is abruptly closed. Below is an by example.
```

jparise (commented):
I'm not sure I understand how this is any different than using a (non-tox) interactive `python` interpreter or a more featured interactive environment like [IPython](https://ipython.org/).

4383:
This is exactly the same, it's just that tox wrap the env, fetch all the requirements and install pymemcache automatically.

jparise:
> This is exactly the same, it's just that tox wrap the env, fetch all the requirements and install pymemcache automatically.

I think it would be the most helpful to phrase the documentation generally in terms of a Python interpreter (perhaps using the three example options: `python`, `ipython`, and `tox`), and then suggesting `tox` as the easiest way to get started for the reasons above.

4383:
Latest changes summary:
- Typo fixed
- Modified sentence to speak about python interpreter

4383:
Thanks for your previous reviews

docs/getting_started.rst
```diff
@@ -174,6 +174,83 @@ the `JsonSerde` from above which is more careful with encodings:
             raise Exception("Unknown serialization format")
 
 
+Interacting with pymemcache
+---------------------------
+
+For testing purpose pymemcache can be used in an interactive mode by using
+the python interpreter or again ipython and tools like tox.
+
+One main avantage of using `tox` to interact with `pymemcache` is that it come
```
jparise, line 183:
```suggestion
One main advantage of using `tox` to interact with `pymemcache` is that it comes
```

docs/getting_started.rst
```diff
@@ -174,6 +174,83 @@ the `JsonSerde` from above which is more careful with encodings:
...
+with it's own virtual environments. It will automatically install pymemcache
+fetch all the needed requirements at run. See the example below:
```
jparise, line 185:
```suggestion
and fetch all the needed requirements at run. See the example below:
```

docs/getting_started.rst
```diff
@@ -174,6 +174,83 @@ the `JsonSerde` from above which is more careful with encodings:
...
+   b'some_value'
+   >>> print(client.get.__doc__)
+        The memcached "get" command, but only for one key, as a convenience.
+        Args:
+          key: str, see class docs for details.
+          default: value that will be returned if the key was not found.
+        Returns:
+          The value for the key, or default if the key wasn't found.
+
+You can instantiate all the class and clients offered by pymemcache.
```
jparise, line 204:
```suggestion
You can instantiate all the classes and clients offered by pymemcache.
```

docs/getting_started.rst
```diff
@@ -174,6 +174,83 @@ the `JsonSerde` from above which is more careful with encodings:
...
+
+Your client will remain open until you decide to close it or until you decide
+to quit your interpreter. It can allow to see what's happen if your server
```
jparise, line 207:
```suggestion
to quit your interpreter. It can allow you to see what's happen if your server
```

docs/getting_started.rst
```diff
@@ -174,6 +174,83 @@ the `JsonSerde` from above which is more careful with encodings:
...
+       return self._fetch_cmd(b'get', [key], False).get(key, default)
+     File "/home/user/pymemcache/pymemcache/client/base.py", line 910, in _fetch_cmd
+       buf, line = _readline(self.sock, buf)
+     File "/home/user/pymemcache/pymemcache/client/base.py", line 1305, in _readline
+       raise MemcacheUnexpectedCloseError()
+   pymemcache.exceptions.MemcacheUnexpectedCloseError
+
+We can see that the connection has been closed.
+
+You can also pass command directly from CLI parameters and get output directly:
```
jparise, line 245:
```suggestion
You can also pass a command directly from CLI parameters and get output directly:
```

jparise (commented):
I left a few wording suggestions. I can also merge your change as-is and do my own follow-up editing pass if you'd prefer that approach.

4383:
> I left a few wording suggestions. I can also merge your change as-is and do my own follow-up editing pass if you'd prefer that approach.

No problem thank you, I addressed your comments with the latest patch set.

PR #224, opened 2019-04-15
Allow to flush data on client reconnect.

4383:
Added "flush_on_reconnect" (defaults to off) to Client and HashClient
which will cause a client that has lost connection to a server and then
reconnects to flush the cache on the reconnect so that it doesn't get old
values from that server.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import errno
 import socket
+import logging
 import six
 
 from pymemcache import pool
@@ -26,6 +27,7 @@ from pymemcache.exceptions import (
     MemcacheUnexpectedCloseError
 )
 
+logger = logging.getLogger(__name__)
 
 RECV_SIZE = 4096
 VALID_STORE_RESULTS = {
@@ -206,7 +208,8 @@ class Client(object):
                  socket_module=socket,
                  key_prefix=b'',
                  default_noreply=True,
-                 allow_unicode_keys=False):
+                 allow_unicode_keys=False,
+                 flush_on_reconnect=False):
         """
         Constructor.
 
@@ -233,6 +236,18 @@ class Client(object):
             store commands (except from cas, incr, and decr, which default to
             False).
           allow_unicode_keys: bool, support unicode (utf8) keys
+          flush_on_reconnect: bool, optional flag which prevents a scenario
+            that can cause stale data to be read: If there's more than one
+            memcached server and the connection to one is interrupted, keys
+            that mapped to that server will get reassigned to another.
+            If the first server comes back, those keys will map to it
+            again. If it still has its data, get()s can read stale data that
+            was overwritten on another server. This flag is off by
+            default for backwards compatibility.
+            CAUTION: Pass True to this flag on a cluster of memcache servers
+            will flush your entire memcache cluster by example if you facing
+            some intermittent connection. be sure to know what you are doing
+            before using this flag.
 
         Notes:
           The constructor does not make a connection to memcached. The first
@@ -254,6 +269,13 @@ class Client(object):
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
         self.allow_unicode_keys = allow_unicode_keys
+        self.flush_on_reconnect = flush_on_reconnect
+        self.flush_on_next_connect = False
+        if self.flush_on_reconnect:
+            logger.warn('You start client with `flush_on_reconnect` activated '
+                        'if your client lose the connection with your '
+                        'memcache server all your data will be flush on the '
+                        'next reconnect to your server!')
 
     def check_key(self, key):
         """Checks key and add key_prefix."""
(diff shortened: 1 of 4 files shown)


pymemcache/client/base.py
```diff
@@ -277,9 +288,14 @@ def _connect(self):
                 sock.setsockopt(self.socket_module.IPPROTO_TCP,
                                 self.socket_module.TCP_NODELAY, 1)
         except Exception:
+            if self.flush_on_reconnect:
+                self.flush_on_next_connect + True
```
```suggestion
                self.flush_on_next_connect = True
```

4383: Good catch!

4383:
Unit tests added

pymemcache/test/test_client.py
```diff
@@ -902,6 +924,49 @@ def test_socket_connect_closes_on_failure(self):
         assert socket_module.sockets[0].connections == []
         assert socket_module.sockets[0].closed
 
+    def test_socket_flush_on_reconnect(self):
+
```
jparise: nit: extra newline

pymemcache/test/test_client.py
```diff
@@ -902,6 +924,49 @@ def test_socket_connect_closes_on_failure(self):
...
+        client._connect()
+        result = client.set(b'key', b'value', noreply=False)
+        assert result is True
+        client.sock.connect_failure = OSError()
+        with pytest.raises(OSError):
+            client._connect()
+        assert client.flush_on_next_connect is True
+
+    def test_socket_not_flush_on_reconnect(self):
+
```
jparise: nit: extra newline

pymemcache/test/test_client.py
```diff
@@ -902,6 +924,49 @@ def test_socket_connect_closes_on_failure(self):
         assert socket_module.sockets[0].connections == []
         assert socket_module.sockets[0].closed
 
+    def test_socket_flush_on_reconnect(self):
+
+        server = ("example.com", 11211)
+        timeout = 2
+        connect_timeout = 3
```
jparise: I'm not sure breaking these values out into variables improves readability. I would probably just write them as keyword values when constructing `Client`:

```python
        client = Client(("example.com", 11211), connect_timeout=3,
                        timeout=2,
                        socket_module=mock_permanent_socket,
                        flush_on_reconnect=True)
```

pymemcache/test/test_client.py
4383: done

pymemcache/test/test_client.py
4383: done

pymemcache/test/test_client.py
4383: done

4383:
I've implemented the flush check

4383:
FYI my latest push correspond to a proper rebase of the master on my branch to avoid wrong git history rebase.

4383:
Latest changes introduces some tests on `get` value after reconnect when flush is disable, we need to retrieve values in this case.

cgordon:
This is a dangerous feature to enable on a memcache cluster. Consider a case in which a client has an intermittent connection to a cluster of memcache servers that are all healthy (and remain healthy). The client could, over a very short period of time, get connection errors to all of the servers, and then subsequently connect successfully to them, issuing a flush command. As a result, over a short period of time (milliseconds) you could end up flushing your entire memcache cluster.

I'm not completely against adding this flag, but it should include a VERY strongly worded warning that it is only to be used in cases where it is acceptable to flush all of the memcache servers in a short period of time.

4383:
> This is a dangerous feature to enable on a memcache cluster. Consider a case in which a client has an intermittent connection to a cluster of memcache servers that are all healthy (and remain healthy). The client could, over a very short period of time, get connection errors to all of the servers, and then subsequently connect successfully to them, issuing a flush command. As a result, over a short period of time (milliseconds) you could end up flushing your entire memcache cluster.
> 
> I'm not completely against adding this flag, but it should include a VERY strongly worded warning that it is only to be used in cases where it is acceptable to flush all of the memcache servers in a short period of time.

Sure I'll add a strong warn related to the usage of this flag.
This flag should be explicitly declared when user initialize client, so I guess we can consider this is an advanced feature (advanced use case) dedicated to users who know that they are doing.
Normal usages are not impacted by this feature.

4383:
Have added warns

4383:
Hey @jparise @cgordon,

Do you have any updates from your side?

cgordon:
I'm just getting back from vacation, so I'll take a look this week, thanks.

4383:
@cgordon ack thanks

pymemcache/client/base.py
```diff
@@ -233,6 +236,18 @@ def __init__(self,
             store commands (except from cas, incr, and decr, which default to
             False).
           allow_unicode_keys: bool, support unicode (utf8) keys
+          flush_on_reconnect: bool, optional flag which prevents a scenario
```
cgordon: I'd prefer something more like this:

flush_on_reconnect: bool, optional flag which causes a `flush_all()` command to be sent anytime the memcache client is disconnected and then reconnected to a memcached server. If you have a cluster of memcached servers that use a consistent hash ring (or a similar approach to routing) this can prevent stale reads of old keys when a server in the ring is disconnected and then reconnected. Note, however, that a client with an intermittent connection to a memcached server can cause it to be repeatedly flushed, and the same is true for a cluster of memcached servers. That makes this a VERY dangerous thing to enable on a memcache cluster.

pymemcache/client/base.py
```diff
@@ -254,6 +269,13 @@ def __init__(self,
         self.key_prefix = key_prefix
         self.default_noreply = default_noreply
         self.allow_unicode_keys = allow_unicode_keys
+        self.flush_on_reconnect = flush_on_reconnect
+        self.flush_on_next_connect = False
+        if self.flush_on_reconnect:
+            logger.warn('You start client with `flush_on_reconnect` activated '
```
cgordon: I'd prefer to note repeatedly warn users, and rather trust that they have read the docs and know their business.

4383:
@cgordon thanks for your reply.
As you want, no problem.

This is a special use case that we could possibly use on openstack, but it does not matter, I can understand your fear.

BTW thanks for your time :)

cgordon:
@4383 Thank you! I appreciate your time, and I'm sorry it took me so long to make up my mind and communicate it.

4383:
@cgordon no problem :)
See you!

PR #225, opened 2019-04-19
Add default_noreply paramter to HashClient

cdonati:
This allows users to specify the noreply behavior upon creation of a HashClient like they are already able to do for Client and PooledClient.


jparise:
Thank you, @cdonati!

PR #232, opened 2019-06-03
add encoding in client functions

JianGuoPinterest:
in order to support remcached operations, I extracted the key helper function, it won't affect the current logic, this _key_helper function will be override in the remcached client implementation.


pymemcache/client/base.py
```diff
@@ -739,6 +739,10 @@ def _raise_errors(self, line, name):
             error = line[line.find(b' ') + 1:]
             raise MemcacheServerError(error)
 
+    # extract the key helper to support extensions
```
jparise: Because we're doing this to support possible extension via subclasses, could you add a little bit more documentation here so that the contract is clearer to end users? I think using a docblock would probably be better than a comment for that purpose, too.

pymemcache/client/base.py
```diff
@@ -304,6 +305,9 @@ def set(self, key, value, expire=0, noreply=None):
                   from the cache, or zero for no expiry (the default).
           noreply: optional bool, True to not wait for the reply (defaults to
                    self.default_noreply).
+          flag: remcached operation flag
```
LINKIWI: The `flag` is not specific to remcached, but rather [one component](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L149) of the memcached protocol's underlying storage command API. Let's update the documentation to reflect this.

pymemcache/client/base.py
```diff
@@ -304,6 +305,9 @@ def set(self, key, value, expire=0, noreply=None):
...
+          encoding: pymemcached use ascii, remcached use utf-8
```
LINKIWI: Again there is no need to reference remcached here as this is a generic, public-facing API. The encoding value here controls how the data is encoded before shipping it on the wire. The fact that remcached chooses `utf-8` is an implementation detail.

pymemcache/client/base.py
```diff
@@ -739,6 +744,22 @@ def _raise_errors(self, line, name):
             error = line[line.find(b' ') + 1:]
             raise MemcacheServerError(error)
 
+    def _key_helper(self, key, remapped_keys, prefixed_keys):
+        """
+        This function is extracted from _fetch_cmd to support
+            future extension for remcached.
+        Remcached will support rich data structure like list/map/set,
```
LINKIWI: This class need not be aware of potential subclass implementations. Same as above, let's remove the reference here.

pymemcache/client/base.py
```diff
@@ -739,6 +744,22 @@ def _raise_errors(self, line, name):
             error = line[line.find(b' ') + 1:]
             raise MemcacheServerError(error)
 
+    def _key_helper(self, key, remapped_keys, prefixed_keys):
```
LINKIWI: While we do need to eventually override this method's logic, by itself it is not particularly useful. It may be more useful to extract the logic under `elif line.startswith(b'VALUE'):` to a separate method. Can you elaborate on the issues you ran into when you tried to do that?

From a cursory glance seems like you can pass in `remapped_keys`, `prefixed_keys`, and `line` as inputs and keep most of the logic intact.

pymemcache/client/base.py
```diff
@@ -293,7 +293,8 @@ def close(self):
             finally:
                 self.sock = None
 
-    def set(self, key, value, expire=0, noreply=None):
+    def set(self, key, value, expire=0, noreply=None, flag=None,
+            encoding='ascii'):
```
jparise: It's a best practice to add new keyword arguments to the end of the signature in case someone has made an assumption about the positional order.

Also, I think this should be spelled `flags` to match the memcached protocol. And should this default to `0` instead of `None`?

pymemcache/client/base.py
```diff
@@ -304,6 +305,9 @@ def set(self, key, value, expire=0, noreply=None):
                   from the cache, or zero for no expiry (the default).
           noreply: optional bool, True to not wait for the reply (defaults to
                    self.default_noreply).
+          flag: memcached operation flag
```
jparise: (Let's swap the order of `flags` and `encoding` here, too.

We should also provide a little more document borrowed from the protocol:

https://github.com/memcached/memcached/blob/78eb7701e0823643d693c1a7a6fd8a0c75db74d8/doc/protocol.txt#L175-L181

Something like:

```
flags: int, arbitrary bit field used for server-specific flags
```

pymemcache/client/base.py
```diff
@@ -739,6 +744,32 @@ def _raise_errors(self, line, name):
             error = line[line.find(b' ') + 1:]
             raise MemcacheServerError(error)
 
+    def _value_helper(self, expect_cas, line, buf, remapped_keys,
+                      prefixed_keys):
+        """
+        This function is abstracted from _fetch_cmd to support different ways
+        of value extraction. In order to use this feature, _value_helper needs
+        to be overriden in the subclass.
+
```
jparise: nit: extra newline

pymemcache/client/base.py
```diff
@@ -755,30 +786,16 @@ def _fetch_cmd(self, name, keys, expect_cas):
             buf = b''
             result = {}
             while True:
+
```
jparise: nit: extra newline

pymemcache/client/base.py
```diff
@@ -755,30 +786,16 @@ def _fetch_cmd(self, name, keys, expect_cas):
...
-                    key = remapped_keys[key]
-
-                    if self.deserializer:
-                        value = self.deserializer(key, value, int(flags))
-
-                    if expect_cas:
-                        result[key] = (value, cas)
-                    else:
-                        result[key] = value
+                    key, value, buf \
```
jparise: `\`-based continuations are discouraged. Please wrap something like this:

```python
key, value, buf = self._value_helper(expect_cas, line, buf,
                                     remapped_keys, prefixed_keys)
```

pymemcache/client/base.py
```diff
@@ -794,7 +811,8 @@ def _fetch_cmd(self, name, keys, expect_cas):
                 return {}
             raise
 
-    def _store_cmd(self, name, values, expire, noreply, cas=None):
+    def _store_cmd(self, name, values, expire, noreply, cas=None, flag=None,
```
jparise: `flag` -> `flags` and move to last argument.

pymemcache/client/base.py
```diff
@@ -803,7 +821,7 @@ def _store_cmd(self, name, values, expire, noreply, cas=None):
             extra += b' ' + cas
         if noreply:
             extra += b' noreply'
-        expire = six.text_type(expire).encode('ascii')
+        expire = six.text_type(expire).encode(encoding)
```
jparise: This looks like a good (but separate) bug fix. Can you submit it as its own Pull Request so we can easily give it its own tests and changelog entry?

(including the related lines below)

pymemcache/client/base.py
```diff
@@ -813,19 +831,23 @@ def _store_cmd(self, name, values, expire, noreply, cas=None):
             if self.serializer:
                 data, flags = self.serializer(key, data)
             else:
-                flags = 0
+                # flag is passed only by remcached operations
```
jparise: Please remove this comment. It's not generally applicable.

pymemcache/client/base.py
```diff
@@ -813,19 +831,23 @@ def _store_cmd(self, name, values, expire, noreply, cas=None):
...
+                if flag:
```
jparise: If we adopt `flags`, then bit of code can just use the `flags` parameter directly.

pymemcache/client/base.py
JianGuoPinterest: remcached only supports text protocol but not binary protocol, in order for the remcached extension to consume pymemcahe, the encoding has to be changed to 'utf-8', that's why I added param encoding. Do you think it's necessary to have a separate PR?

pymemcache/client/base.py
jparise: Please ignore the first part of my comment. I mistakenly thought that `encoding` already existed here.

pymemcache/client/base.py
jparise: Got it. I _do_ think it would be helpful to break this into three smaller (sequential) PRs:

1. The `encoding` addition.
2. The `flags` addition.
3. The "value helper".

JianGuoPinterest: totally! Will do!

pymemcache/client/base.py
```diff
@@ -304,6 +304,7 @@ def set(self, key, value, expire=0, noreply=None):
                   from the cache, or zero for no expiry (the default).
           noreply: optional bool, True to not wait for the reply (defaults to
                    self.default_noreply).
+          encoding: determines how the data is encoded(binary or text).
```
```suggestion
          encoding: optional str, controls data encoding (defaults to 'ascii').
```

LINKIWI:
The smaller-scope change looks pretty good to me. As jparise said let's add a test case for a non-default encoding, after which this will probably be good to merge.

JianGuoPinterest:
> Should we also add `encoding` to `set_many()`?
 
On remcached side, we do not consume the set_many() function, so logically we don't have to.

jparise:
> > Should we also add `encoding` to `set_many()`?
> 
> On remcached side, we do not consume the set_many() function, so logically we don't have to.

I think we should add it to `set_many()` for consistency, especially because it just calls through to the existing support in `_store_cmd()`.

pymemcache/client/base.py
```diff
@@ -794,7 +796,8 @@ def _fetch_cmd(self, name, keys, expect_cas):
                 return {}
             raise
 
-    def _store_cmd(self, name, values, expire, noreply, cas=None):
+    def _store_cmd(self, name, values, expire, noreply, cas=None,
+                   encoding='ascii'):
```
jparise: Because `_store_cmd()` is an internal method, I think we should move `encoding` just before `cas` and _not_ give it a default value (so no keyword argument) because we're always providing a value for it from the public methods.

JianGuoPinterest: _store_cmd() is being called by add()/ append()/ cas()/ prepend()/ replace()/ set()/ set_many(), from what you mentioned above, in order to make it consistency, should we add encoding for all these functions?

jparise: > _store_cmd() is being called by add()/ append()/ cas()/ prepend()/ replace()/ set()/ set_many(), from what you mentioned above, in order to make it consistency, should we add encoding for all these functions?

Eek, you're right. Should `encoding` be a Client-wide setting (set in `__init__()`)? Or might we vary it based on each `set` operation?

JianGuoPinterest: I'm pretty sure across remcached usage, we only use text protocol which is 'utf-8'. if we add encoding to `__init__()`, then in set and _store_cmd function, we won't need to add parameter encoding. Since the default encoding will be 'ascii', I don't think adding encoding into `__init__()` could cause any trouble for the current usage. If you think it's a good solution, i'll give it a try.

jparise: > I'm pretty sure across remcached usage, we only use text protocol which is 'utf-8'. if we add encoding to `__init__()`, then in set and _store_cmd function, we won't need to add parameter encoding. Since the default encoding will be 'ascii', I don't think adding encoding into `__init__()` could cause any trouble for the current usage. If you think it's a good solution, i'll give it a try.

Assuming we're only going to apply this to value (and not key) encoding, I think setting this in `__init__()` will work well.

Keys are expected to be ASCII-encoded so that shouldn't be configurable: https://github.com/memcached/memcached/blob/78eb7701e0823643d693c1a7a6fd8a0c75db74d8/doc/protocol.txt#L612-L615

JianGuoPinterest: yes, it's only for values. I'll work on it now.

pymemcache/test/test_client.py
```diff
@@ -131,9 +131,21 @@ def test_set_success(self):
...
+        # unit test for encoding passed in __init__()
+        client = self.make_client([b'STORED\r\n'], encoding='utf-8')
+        result = client.set(b'key', b'value', noreply=False)
+        assert result is True
+
     def test_set_future(self):
         client = self.make_client([b'STORED\r\n'])
-        result = client.set(newbytes(b'key'), newbytes(b'value'), noreply=False)
+        result = client.set(newbytes(b'key'), newbytes(b'value'),
+                            noreply=False)
```
jparise: I think, if you move `noreply` back to the pervious line as it was before, we'll get a cleaner diff here (and there isn't a reason to break this line here).

PR #235, opened 2019-06-09
add flags in client functions

JianGuoPinterest:
added flags in client functions to provide support for extra memcached operations.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -296,7 +296,7 @@ class Client(object):
             finally:
                 self.sock = None
 
-    def set(self, key, value, expire=0, noreply=None, flags=0):
+    def set(self, key, value, expire=0, noreply=None, flags=None):
         """
         The memcached "set" command.
 
@@ -320,7 +320,7 @@ class Client(object):
         return self._store_cmd(b'set', {key: value}, expire, noreply,
                                flags=flags)[key]
 
-    def set_many(self, values, expire=0, noreply=None, flags=0):
+    def set_many(self, values, expire=0, noreply=None, flags=None):
         """
         A convenience function for setting multiple values.
 
@@ -345,7 +345,7 @@ class Client(object):
 
     set_multi = set_many
 
-    def add(self, key, value, expire=0, noreply=None, flags=0):
+    def add(self, key, value, expire=0, noreply=None, flags=None):
         """
         The memcached "add" command.
 
@@ -369,7 +369,7 @@ class Client(object):
         return self._store_cmd(b'add', {key: value}, expire, noreply,
                                flags=flags)[key]
 
-    def replace(self, key, value, expire=0, noreply=None, flags=0):
+    def replace(self, key, value, expire=0, noreply=None, flags=None):
         """
         The memcached "replace" command.
 
@@ -393,7 +393,7 @@ class Client(object):
         return self._store_cmd(b'replace', {key: value}, expire, noreply,
                                flags=flags)[key]
 
-    def append(self, key, value, expire=0, noreply=None, flags=0):
+    def append(self, key, value, expire=0, noreply=None, flags=None):
         """
         The memcached "append" command.
 
@@ -415,7 +415,7 @@ class Client(object):
         return self._store_cmd(b'append', {key: value}, expire, noreply,
                                flags=flags)[key]
 
-    def prepend(self, key, value, expire=0, noreply=None, flags=0):
+    def prepend(self, key, value, expire=0, noreply=None, flags=None):
         """
         The memcached "prepend" command.
 
@@ -437,7 +437,7 @@ class Client(object):
         return self._store_cmd(b'prepend', {key: value}, expire, noreply,
                                flags=flags)[key]
 
-    def cas(self, key, value, cas, expire=0, noreply=False, flags=0):
+    def cas(self, key, value, cas, expire=0, noreply=False, flags=None):
         """
         The memcached "cas" command.
 
@@ -817,7 +817,7 @@ class Client(object):
                 return {}
             raise
 
-    def _store_cmd(self, name, values, expire, noreply, flags=0, cas=None):
+    def _store_cmd(self, name, values, expire, noreply, flags=None, cas=None):
         cmds = []
         keys = []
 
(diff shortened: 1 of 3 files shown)


pymemcache/client/base.py
```diff
@@ -815,8 +818,6 @@ def _store_cmd(self, name, values, expire, noreply, cas=None):
             key = self.check_key(key)
             if self.serializer:
                 data, flags = self.serializer(key, data)
```
jparise: It looks like this local `flags` variable will shadow the `flags` argument. How do they relate?

pymemcache/client/base.py
```diff
@@ -307,6 +307,7 @@ def set(self, key, value, expire=0, noreply=None):
                   from the cache, or zero for no expiry (the default).
           noreply: optional bool, True to not wait for the reply (defaults to
                    self.default_noreply).
+          flags: int, arbitrary bit field used for server-specific flags
```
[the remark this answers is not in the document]
JianGuoPinterest: the original logic is, if a serializer is specified when constructing the client, it will generate the flags using the serializer, otherwise flags will be set to 0 as default. 
Current logic won't change that part of logic, it will still generate flags if serializer is specified while constructing the client, or default to 0 if not specified. 
For the extension logic, we do not specify serializer, so it would not use that piece of logic. We will pass flags for the set function. 
This will not affect the current usage of set function, the code worked before could still work with the new logic.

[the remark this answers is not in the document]
jparise: (I think this was a reply to the latter comment, so I'll move my reply there.)

pymemcache/client/base.py
jparise: I think this creates an unexpected case where `flags` is passed in but is overridden by the serializer. I can think of two improvements:

1. Because `flags` is a bit field, we could `OR` the two values together (the `flags` argument plus these `serializer_flags`) to form the final value send to the server. I don't know if that would always be the right thing to do, however, because we can't know in advance how the server and serializer are going to use the flags.
2. We could raise an exception when both `flags` and `self.serializer` are provided because they're incompatible.

JianGuoPinterest: One more solution would be while using `data, flags = self.serializer(key, data)`, we could check if the `flags` is being passed, if `flags!=0`, we could use the flags generated by the `serializer`. Also, we could add more comments to tell the consumers about the logic of `serializer` and `flags` so they could be aware of the conflict passing `serializer` and `flags` at the same time.

jparise: That sounds like a variant of (1) above (where we consider both input values and make a decision of which bits we want to send to the server). It has the same potential downsides, but given that we're baselining on `0` being "safe", it could be a good compromise.

pymemcache/client/base.py
```diff
@@ -814,9 +818,10 @@ def _store_cmd(self, name, values, expire, noreply, cas=None):
 
             key = self.check_key(key)
             if self.serializer:
-                data, flags = self.serializer(key, data)
-            else:
-                flags = 0
+                if flags == 0:
```
jparise: How about:

```python
if self.serializer:
    data, serializer_flags = self.serializer(key, data)
    # Use the serializer's flags when `flags` haven't been specified.
    if flags == 0:
        flags = serializer_flags
```

JianGuoPinterest: looks good, will change the code base on it.

pymemcache/client/base.py
```diff
@@ -307,6 +307,8 @@ def set(self, key, value, expire=0, noreply=None):
                   from the cache, or zero for no expiry (the default).
           noreply: optional bool, True to not wait for the reply (defaults to
                    self.default_noreply).
+          flags: optional int, arbitrary bit field used for server-specific
+                flags
```
jparise: Should we write something here about the new behavior when a serializer is present?

pymemcache/client/base.py
```diff
@@ -315,7 +317,8 @@ def set(self, key, value, expire=0, noreply=None):
         """
         if noreply is None:
             noreply = self.default_noreply
-        return self._store_cmd(b'set', {key: value}, expire, noreply)[key]
+        return self._store_cmd(b'set', {key: value}, expire, noreply,
+                               flags=flags)[key]
 
     def set_many(self, values, expire=0, noreply=None):
```
jparise: Does it make sense for `set_many` to also accept `flags`?

JianGuoPinterest: so `_store_cmd()` is being called by `add()`/ `append()`/ `cas()`/` prepend()`/` replace()`/ `set()`/ `set_many()`. I mean I can add flags to all those functions, but if there are no use cases consuming flags from these functions, adding flags might make it more confusing for the consumers.

jparise: I think it would make sense to add them there in the long-term, but I'm comfortable waiting until people need them given that this is still a new capability.

pymemcache/client/base.py
JianGuoPinterest: Which of these two params should have higher priority? In case they passed serializer and flags together, flags is not 0 generated from serialized, should we just ignore the flags passed by set function? This does not sound right to me.

pymemcache/client/base.py
JianGuoPinterest: I think put the comments in the _store_cmd() would be better since we might add flags for other functions as well.

pymemcache/client/base.py
jparise: If we're going to choose one or the other, I think we should prefer the explicit parameter passed to the function.

But this is also why I was thinking we should consider merging the two bit fields (`flags | serialized_flags`) and sending that value to the server.

JianGuoPinterest: for the bit operation, my concern would be, system might work fine if we just pass flags or serialized_flags, but if we do the bitwise operation, it could end up with exception. I would vote for using the flags passed by set cause that one should be the one we should use.

jparise: I think that's fine. If someone is setting that value, they hopefully know what they're doing based on how their environment (memcached and chosen serializer) is configured.

jparise: It does make me think that we might want to default `flags=None` in the signature, however, because it's possible someone might want to "force" `flags=0` to override the serializer's behavior. What do you think?

JianGuoPinterest: the default behavior is `flags=0`. what you mentioned is a really corner case. I highly doubt this case would happen. but setting `flags=None` does make sense.  I will modify the code.

Issue #239, opened 2019-07-23
Docs should note that `nodelay=True` can lead to spurious "MemcacheUnknownCommandError"


sirosen:
Porting code which had been using python-memcached to pymemcache, I uncovered an issue in which I was setting a fractional `expire`. (`python-memcached` had hidden this from me entirely, but the bug was there even before.)
However, this was hard to debug because (1) the fractional expiration was never passed by our testsuite (2) I had set `noreply=True` on all `Client.set()` calls, hiding the erroneous expiration.

The result was that I was getting `pymemcache.exceptions.MemcacheUnknownCommandError: b'get'` which was "obviously wrong" and could not reproduce with various tests.
Digging into pymemcache, I saw that these errors are triggered by reading `ERROR` immediately after sending a command, and guessed that an error from a prior command was left in the channel, being read when the `Client.get()` was evaluated.

I'm not sure if there's a better way to handle
```python
from pymemcache.client.base import Client
client = Client(("localhost", 11211))
client.set("foo", "bar", expire=1.5, noreply=True)
client.get("foo")
```

than what's currently being done? It seems reasonable, but I think a documentation warning could save others this same mistake and hassle.

Although this particular case could be handled by checking if `expire` isn't a int, I don't think that's particularly important.

cgordon:
Hi Stephen,

Thank you for the excellent bug report! I was surprised by this behavior, so I went and read back through the protocol documentation:

https://github.com/memcached/memcached/blob/master/doc/protocol.txt

It says this:

> "noreply" optional parameter instructs the server to not send the
>  reply.  NOTE: if the request line is malformed, the server can't
>  parse "noreply" option reliably.  In this case it may send the error
>  to the client, and not reading it on the client side will break
> things.  Client should construct only valid requests.

In light of that, I think the right answer is to validate all the arguments passed and fail immediately if any of them would cause us to construct an invalid request. In this case, that means checking that the value of expire is an integer before serializing it to the request string.

Would you like to send us a PR for that?

Thank you,
Charles

sirosen:
Sure, I'll do that momentarily.

I was thinking that catching this specific kind of malformed request was less important than having some kind of note in the docs that `noreply=True` can "defer errors until the next call to memcache". What do you think of also making a doc change of some kind? I'll keep it separate if I open a PR so that the issues can be reviewed independently.

jparise:
I think that's also a good idea, @sirosen. Thanks!

PR #240, opened 2019-07-24
Add check that `expire` is an integer

sirosen:
Part of #239

If a caller tries to send a fractional expiration time, that should be caught and rejected with a MemcacheIllegalInputError. Otherwise, calls with `noreply=True` could fail to parse, but the client won't pick up on this until the next call to memcache which reads the reply.

The case we're seeking to avoid is this:

```python
client = Client(("localhost", 11211))
client.set("foo", "bar", expire=1.5, noreply=True)
client.get("foo")  # triggers MemcacheUnknownCommandError
```

In addition to allowing for values which are `int` (or `long` on py2), this implementation allows for `expire` values to be a text type (str on py2, bytes or str on py3), so long as `int(expire)` correctly converts it. If `int(expire)` fails (ValueError), that indicates that the input could not be treated as an int and is a failure.

Applies `Client._store_cmd`, which covers a class of commands, and `Client.touch`, which appears to be the only command with a non-shared implementation that includes `expire` as a parameter.

---

I tried to write the test using `pytest.mark.parametrize`, but something about the way the testsuite is structured makes that not work (or at least, not work in the way that I'm used to). I'm happy to rewrite the test in a form which is more fitting with the idioms of the current testsuite, but I'd need some direction on that.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -113,7 +113,8 @@ def _check_integer(value, name, encoding):
     """Check that a value is an integer and encode it as a binary string"""
     if isinstance(value, six.integer_types):  # includes "long" on py2
         return six.text_type(value).encode(encoding)
-    elif isinstance(value, six.text_type):
+    # TODO: drop this behavior in a future version of pymemcached
+    elif isinstance(value, VALID_STRING_TYPES):
         try:
             int(value)
             return six.text_type(value).encode(encoding)
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -109,6 +109,22 @@ def _check_key(key, allow_unicode_keys, key_prefix=b''):
     return key
 
 
+def _check_integer(value, name, encoding):
+    """Check that a value is an integer and encode it as a binary string"""
+    if isinstance(value, six.integer_types):  # includes "long" on py2
+        return six.text_type(value).encode(encoding)
+    elif isinstance(value, six.text_type):
+        try:
+            int(value)
+            return six.text_type(value).encode(encoding)
+        except ValueError:
+            pass  # fall-through to error below
+
+    raise MemcacheIllegalInputError(
+        '{} must be integer, got bad value: {}'.format(name, value)
+    )
+
+
 class Client(object):
     """
     A client for a single memcached server.
@@ -648,9 +664,8 @@ class Client(object):
         if noreply is None:
             noreply = self.default_noreply
         key = self.check_key(key)
-        cmd = (
-            b'touch ' + key + b' ' + six.text_type(expire).encode(self.encoding)
-        )
+        expire = _check_integer(expire, "expire", self.encoding)
+        cmd = b'touch ' + key + b' ' + expire
         if noreply:
             cmd += b' noreply'
         cmd += b'\r\n'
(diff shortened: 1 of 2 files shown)


sirosen:
I'm fine (even happy) with requiring it to match `six.integer_types`. I just know that this currently does `six.text_type(expire)`, so if you pass `expire="1"`, that's okay (if a bit weird):
https://github.com/pinterest/pymemcache/blob/c42453c336fece50a123f410554df0bb55b21859/pymemcache/client/base.py#L839

I'd be worried that changing this could break callers who are currently passing strings. Even if it's weird, I know I showed up on this project because I was passing invalid inputs, so I'm sympathetic. 😉

Also, I just noticed that `Client.touch` uses `'ascii'` rather than `self.encoding`... Separate PR incoming.

jparise:
> I'm fine (even happy) with requiring it to match `six.integer_types`. I just know that this currently does `six.text_type(expire)`, so if you pass `expire="1"`, that's okay (if a bit weird)

Ack, you're right! I missed that we're already supporting those class of inputs. I think it would be nice to drop that support, but that probably shouldn't be done in a minor version.

pymemcache/client/base.py
```diff
@@ -109,6 +109,22 @@ def _check_key(key, allow_unicode_keys, key_prefix=b''):
     return key
 
 
+def _check_integer(value, name):
```
jparise: What do you think about this function returning the value that we ultimately send to the server (instead of just checking its type)? That could save us a least a little work when this function detects we've been given a text input value.

sirosen: Yeah, I'll make that edit and mark this resolved when it's done.

sirosen: I ended up rebasing, since this conflicts mildly with #242. But I've changed `_check_integer` to do the encoding work.

pymemcache/client/base.py
```diff
@@ -109,6 +109,22 @@ def _check_key(key, allow_unicode_keys, key_prefix=b''):
...
 
 
+def _check_integer(value, name, encoding):
+    """Check that a value is an integer and encode it as a binary string"""
+    if isinstance(value, six.integer_types):  # includes "long" on py2
+        return six.text_type(value).encode(encoding)
+    elif isinstance(value, six.text_type):
+        try:
+            int(value)
+            return six.text_type(value).encode(encoding)
```
jparise: Because `value` is already a `six.text_type`, could we just do:

```suggestion
            return value.encode(encoding)
```

sirosen: Agh! You're right, but this makes me realize that I've introduced a pernicious bug.
I was thinking about someone passing `expire="1"`, but they could also pass `expire=b"1"` today, just as well. Under my changes `b"1"` would be rejected as invalid (but prior to this PR, it was allowed).

I doubt anyone would do _either_ of these on purpose, but I could imagine loading config or some other data and failing to properly convert it. And in that case, `b"1"` doesn't seem much less likely than `"1"`. I think this should be checking for `six.string_types` above, and still doing `six.text_type(value)` here.

Aside: this makes me think that dropping support for `str` / `bytes` / `unicode` inputs for `expire`, `delay`, and other integer params is well worth doing.

jparise: That makes sense, and I _do_ think it would be nice to drop type conversion for these integer parameters. I see in our documentation comments that we call those fields out as `int` already.

Maybe this is something we can do for the next 2.x release given it's only been semi-supported thus far.

sirosen: I'm going to open a competing/alternate PR which enforces that int parameters are really integers. All of my fretting about "backwards compatibility" might be to the detriment of the project, long term, and I don't want to be saddling people with something subpar.

sirosen:
#243 does this, but in a way that seems better for the project long-term and has people's buy-in. I'm closing as a duplicate.

PR #241, opened 2019-07-24
Add noreply warning to getting_started doc

### LINKED ISSUE #239: Docs should note that `nodelay=True` can lead to spurious "MemcacheUnknownCommandError"

sirosen:
resolves #239

Add a warning to the Best Practices section of the Getting Started doc which explains the interaction between `noreply=True` and error handling.

Also, fix a very minor typo, ``"noreply`` in a nearby line.

I understand that "WARNING" boxes are very shouty and attention-grabby, and many people are shy about using them. However, (1) in the readthedocs theme, they're orange, not red, and less harsh, and (2) this seems genuinely warning-worthy to me.

I thought for a moment about how to fit this into the client documentation, but it doesn't fit well anywhere in there, IMO.

For ease of review, here's what it looks like when rendered:

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -128,7 +128,7 @@ Best Practices
    your process when memcached is slow. You might also want to enable the
    ``no_delay`` option, which sets the TCP_NODELAY flag on the connection's
    socket.
- - Use the ``noreply`` flag for a significant performance boost. The ``"noreply``
+ - Use the ``noreply`` flag for a significant performance boost. The ``noreply``
    flag is enabled by default for "set", "add", "replace", "append", "prepend",
    and "delete". It is disabled by default for "cas", "incr" and "decr". It
    obviously doesn't apply to any get calls.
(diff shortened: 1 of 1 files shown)


jparise:
Thanks!

PR #243, opened 2019-07-25
Check integer input values to Client methods

sirosen:
After the conversation on #240 , I'm starting to change my mind on whether or not it's even worthwhile to continue to support callers using `expire="1"` and the like. That usage is not documented, and supporting it doesn't really make things that much better for callers.

If a caller tries to send
- a fractional expiration time
- a string for the "incr" command
- delay=None (instead of delay=0)

The pymemcache base client will now catch these usage mistakes and throw a MemcacheIllegalInputError.

Although there are existing valid (but undocumented) usages of the current string conversion behavior, the docs state that these values are supposed to be ints. Now if a non-int value is used, it will immediately trigger an error.

This behavior is superior to the prior behavior especially in the case where a command is used with `noreply=True`, as described in #240 (the commit message here has it too). 

Applies to all commands using `expire`, `delay`, incr and decr inputs, and memlimit.

Add a note to the changelog which puts positive spin on this: pymemcache is now validating inputs which it didn't used to validate.

closes #240

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,6 +1,12 @@
 Change Log
 ==========
 
+New in version 3.0.0 (unreleased)
+---------------------------------
+
+* Validate integer inputs for ``expire``, ``delay``, ``incr``, ``decr``, and
+  ``memlimit`` -- non-integer values now cause ``MemcacheIllegalInputError``s
+
 New in version 2.2.0
 --------------------
 * Drop official support for Python 3.4.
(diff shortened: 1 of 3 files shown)


jparise:
I also prefer this approach. Here's my proposed way forward:

1. Release version 2.2.0 soon (hopefully this week).
2. Bump the project to 3.0 and land this as the first comment on that new major release line.

jparise:
@sirosen I've just release version 2.2.0 so please update your ChangeLog entry accordingly. Thanks!

sirosen:
I've rebased, updating the commit message slightly and altering the changelog entry to be under `3.0.0 (unreleased)`.

jparise:
Thanks again, @sirosen! I've merge this as the first commit in the 3.0 line.

PR #245, opened 2019-07-30
Change serialization interface to be an object

### LINKED ISSUE #56: Change the serializer API to be a single class
I don't think taking `serializer` and `deserializer` as arguments is a clean API because they really should be treated as a protocol.   I can't imagine a time you would provide a `serializer` and not have to provide a `deserializer` for it to successfully work.

I suggest the following:

```python

class JsonSerializer:
     def serialize(self, key, value):
         if type(value) == str:
             return value, 1
         return json.dumps(value), 2

    def deserialize(self, key, value, flags):
        if flags == 1:
            return value
        if flags == 2:
            return json.loads(value)
        raise Exception("Unknown serialization format")

client = Client(('localhost', 11211), serializer=JsonSerializer())
```

sirosen:
Rather than passing separate serialization and deserialization methods to a pymemcache client, pass an object implementing a very simple two-method interface.

This is a rather significant breaking change and should be part of an x.0.0 major release.

Resolves #56

As suggested in that issue, this is a cleaner interface, as there's no sensible context in which you would provide only one of these two methods and it should therefore be thought of as a serialization/deserialization protocol.

Also adds a note to the documentation's Best Practices list that you should use the built-in serializer object unless you have a reason to do otherwise.

I intentionally phrased things in the docs to suggest that you _do not_ need to subclass `PythonMemcacheSerializer`. Just implement the same two-method interface which is already in use.
This also makes it pretty trivial to port any pymemcache 2.x code to 3.x if this is accepted, as you can just define a shim object, a la

```python
def old_serializer(key, value):
    ...
def old_deserializer(key, value, flags):
    ...
class ShimSerializer(object):
    def serialize(self, *args, **kwargs):
        return old_serializer(*args, **kwargs)
    def deserialize(self, *args, **kwargs):
        return old_deserializer(*args, **kwargs)
```

---

@jparise, I noticed you created a 3.0 milestone and added a couple of issues to it. This inspired me to take a crack at #56, which seems like a nice improvement to the interfaces presented by Pymemcache,  but obviously needs to be a major release.


jparise:
We could in theory support both forms by naming this new parameter `serde` and internally providing a wrapping object if the legacy `serializer` and `deserializer` were provided.

sirosen:
After some thought, I don't really like the idea of doing that *long-term*. #56 is about cleaning up the API for Client, and adding more args makes it messier. Unless we agree that the compatible behavior should be dropped in a future major release, I would rather not make this change at all. Does that seem reasonable?

I acknowledge that changing the public API for one of the main interfaces of pymemcache is a "very breaky" change -- almost every user of the library would be impacted.

I'm going to add a commit which
1. Renames this param from `serializer` to `serde`
2. Makes the change to add a wrapping object if both `serializer` and `deserializer` are provided
3. Raises a `DeprecationWarning` in this usage, stating that support for this behavior will be dropped in pymemcache version 4
4. Renames the objects in `pymemcache.serde` to preserve the old public function names, so that prior usages of those will still work

Does that sound good?

sirosen:
I've gotten this work done and passing tests for me, but I need to get on a flight. It needs more testing, since there are now multiple usages supported. So please don't merge in this state -- it's a work in progress -- but I wanted to get the new version out there for review.

jogo:
I am weary of making such a major breaking change just to clean up an API for the sake of making it cleaner.

sirosen:
@jogo, I changed the deprecation warning to not commit to removing the supported methods in v4. Does that make it more palatable?

Personally, I don't mind the current interface that much, but given that #56 has been left open, it seemed to me like it had a kind of "tacit approval" from the pymemcache owners. I only went after it because it seems 3.0 is around the corner.

I also tweaked things a little to use a "no-op serializer" when nothing is specified. This doesn't change semantics at all, but it lets any operations that want to serialize or deserialize data skip the `if self.serializer` checks that exist today -- just call `self.serde.serialize(...)` unconditionally.
I think this makes things easier to read, but if it's not liked or wanted, I can remove the commit.

pymemcache/client/base.py
```diff
@@ -109,6 +111,28 @@ def _check_key(key, allow_unicode_keys, key_prefix=b''):
...
+
+def _load_serde(serde, serializer, deserializer):
+    """helper for handling compatible "serde" object loading"""
+    if serde:
+        return serde
+    elif serializer and deserializer:
+        warnings.warn(
+            "pymemcache.base.Client support for the serializer and "
+            "deserializer functions is deprecated and will be removed in "
+            "a future version of pymemcache. Use the 'serde' object instead.",
```
jogo: I don't want to commit to saying will be removed in a future version I would rather leave it as deprecated but have no plans to remove it unless we need to. I don't want to force extra work for existing users if possible

sirosen:
I've done a rebase to resolve several issues.
1. Remove the deprecation warning entirely.
2. Resolve conflicts with `master`
3. Fix the fact that these changes were causing the tests in 7ad74da1 to start failing

A naive rebase resulted in some tests which passed a serializer but no deserializer failing -- the serializer was being discarded in these cases, on the assumption that "serializer without deserializer" was not valid. Even if that's logical or "should be true", it's not necessarily the case today, and there may be valid clients out there passing a serializer without a deserializer. Those should still be supported in this version.

In order to handle this best, I've combined the "NoOpSerde" object with the wrapper for serializer/deserializer functions. There's now just one class, `pymemcache.serde.LegacyWrappingSerde` which implements both default behaviors and allows either one (or both) to be overridden.

This has the potentially "ugly" side effect that now creating a client without `serde`, `serializer`, or `deserializer` will cause `LegacyWrappingSerde` to be used internally, and this may appear in stacktraces and other debugging contexts (vs. `PythonMemcacheNoOpSerde`, which was maybe a little nicer). I think that's fine in exchange for more maintainable library code.

ChangeLog.rst
```diff
@@ -15,6 +15,14 @@ New in version 2.2.1
 --------------------
 * Fix ``flags`` when setting multiple differently-typed values at once.
 
+New in version 3.0.0 (unreleased)
+---------------------------------
+
+* The serialization API has been reworked. Instead of consuming a serializer
```
jparise: You'll need to manually move this new entry up to the (new) 3.0.0 section.

docs/getting_started.rst
```diff
@@ -140,6 +151,9 @@ Best Practices
    errors, from killing your web requests. Do not use this flag if you need to
    know about errors from memcache, and make sure you have some other way to
    detect memcache server failures.
+ - Unless you have a known reason to do otherwise, use the provided serializer
+   in `pymemcache.serde.python_memcache_pickle_serde` for any de/serialization
+   of objects
```
jparise:
```suggestion
   of objects.
```

pymemcache/client/base.py
```diff
@@ -983,8 +988,8 @@ def __init__(self,
                  allow_unicode_keys=False,
                  encoding='ascii'):
         self.server = server
-        self.serializer = serializer
-        self.deserializer = deserializer
+        self.serde = (serde if serde else
+                      LegacyWrappingSerde(serializer, deserializer))
```
```python
self.serde = serde or LegacyWrappingSerde(serializer, deserializer)
```

pymemcache/serde.py
```diff
@@ -105,3 +105,53 @@ def python_memcache_deserializer(key, value, flags):
             return None
 
     return value
+
+
+class PythonMemcachePickleSerde(object):
```
jparise: It seems like we should name this just `PickleSerde`. What do you think?

```suggestion
class PickleSerde(object):
```

pymemcache/test/test_integration.py
```diff
@@ -26,9 +26,7 @@
     MemcacheClientError
 )
 from pymemcache.serde import (
-    get_python_memcache_serializer,
-    python_memcache_serializer,
-    python_memcache_deserializer
+    PythonMemcachePickleSerde, python_memcache_pickle_serde
```
jparise: Let's keep the "one name per line" format we had before.

sirosen: Sure. It reminds me that I meant to suggest maybe using `black`.

I feel like the parts of my brain which pay attention to formatting minutiae have been mostly "turned off" since I started using it.
But definitely, I'll fix to be consistent with what's here.

pymemcache/serde.py
sirosen: I'm a huge 👍 on the more concise name. I think I was just trying to stay consistent with the things which are already in this module.

sirosen:
@jparise, thanks for the review. I've done my last set of changes in a new commit rather than another rebase to make it easier to review the results.

PR #261, opened 2019-12-17
Add prepend and append support to MockMemcacheClient

jogo:
So that MockMemcacheClient is more complete and can be used for testing
prepend and append cases.

### DIFF
--- a/pymemcache/test/utils.py
+++ b/pymemcache/test/utils.py
@@ -138,6 +138,30 @@ class MockMemcacheClient(object):
             self.delete(key, noreply)
         return True
 
+    def prepend(self, key, value, expire=0, noreply=None, flags=None):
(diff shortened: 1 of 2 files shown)


pymemcache/test/utils.py
```diff
@@ -138,6 +138,18 @@ def delete_many(self, keys, noreply=True):
...
+    def prepend(self, key, value, expire=0, noreply=None, flags=None):
+        current = self.get(key)
+        if current is not None:
+            self.set(key, value + current)
+        return True
+
+    def append(self, key, value, expire=0, noreply=None, flags=None):
+        current = self.get(key)
+        if current is not None:
+            self.set(key, value + current)
```
jparise: Should this be `current + value`?

PR #266, opened 2020-01-07
Adding clear method to MockMemcacheClient

FerasAlazzeh:
Adding a `clear` method to `MockMemcacheClient` as a way to reset/clear the mock cache.

### DIFF
--- a/pymemcache/test/utils.py
+++ b/pymemcache/test/utils.py
@@ -43,6 +43,10 @@ class MockMemcacheClient(object):
         self.ignore_exc = ignore_exc
         self.encoding = encoding
 
+    def clear(self):
+        """Method used to clear/reset mock cache"""
+        self._contents.clear()
+
     def get(self, key, default=None):
         if not self.allow_unicode_keys:
             if isinstance(key, six.string_types):


pymemcache/test/utils.py
```diff
@@ -43,6 +43,9 @@ def __init__(self,
         self.ignore_exc = ignore_exc
         self.encoding = encoding
 
+    def clear(self):
+        self._contents.clear()
```
jparise: Let's add a docblock here because this isn't a standard memcache command.

FerasAlazzeh: Sounds good. Let me know if I should add more for ex. (this method is used strictly for testing to increase convenience).

pymemcache/test/utils.py
```diff
@@ -43,6 +43,10 @@ def __init__(self,
         self.ignore_exc = ignore_exc
         self.encoding = encoding
 
+    def clear(self):
+        '''Method used to clear/reset mock cache'''
```
jparise: We use `"` for docblocks elsewhere in the code. Let's stay consistent.

```suggestion
        """Method used to clear/reset mock cache"""
```

jparise:
Thanks!

PR #271, opened 2020-03-18
HashClient pass encoding servers

seahawks8:
pass the encoding from the hashed client to the added servers


seahawks8:
> Thanks, @seahawks8! Could you include a small test case for this, too? It should be easy to add and will help ensure this behavior doesn't regress.

No problem will do later today. I also have a pickle update for deserialization. I will send as a separate pull request

PR #275, opened 2020-03-30
Fix corner case with dead server coming back alive

scoiatael:
Hey, 

I haven't opened an issue for that as I believe it's easier to show failing tests.

Looks like `clients/hash` behaves incorrectly when dead server comes back alive - it's never removed from `dead_clients` and thus retried indefinitely. 

Failing test and fix are visible in 35ee9de, rest of PR simply adds some test coverage and refactors existing code a little.

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -141,7 +141,7 @@ class HashClient(object):
 
     def _get_client(self, key):
         check_key_helper(key, self.allow_unicode_keys, self.key_prefix)
-        if len(self._dead_clients) > 0:
+        if self._dead_clients:
             self._retry_dead()
 
         server = self.hasher.get_node(key)
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -127,14 +127,18 @@ class HashClient(object):
         # we have dead clients and we have reached the
         # timeout retry
         if current_time - ldc > self.dead_timeout:
+            candidates = []
             for server, dead_time in self._dead_clients.items():
                 if current_time - dead_time > self.dead_timeout:
-                    logger.debug(
-                        'bringing server back into rotation %s',
-                        server
-                    )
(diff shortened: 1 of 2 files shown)


pymemcache/client/hash.py
```diff
@@ -121,22 +121,29 @@ def remove_server(self, server, port):
...
+                    'bringing server back into rotation %s',
+                    server
+                )
+                self.add_server(*server)
+                del self._dead_clients[server]
+            self._last_dead_check_time = current_time
+
     def _get_client(self, key):
         check_key_helper(key, self.allow_unicode_keys, self.key_prefix)
         if len(self._dead_clients) > 0:
```
jparise: While we're here, this line could be simplified:

```suggestion
        if self._dead_clients):
```

pymemcache/client/hash.py
```diff
@@ -121,22 +121,29 @@ def remove_server(self, server, port):
         key = '%s:%s' % (server, port)
         self.hasher.remove_node(key)
 
+    def _retry_dead(self):
+        current_time = time.time()
+        ldc = self._last_dead_check_time
+        # we have dead clients and we have reached the
+        # timeout retry
```
jparise: Note that the guarantee that we have dead clients is now external to this function.

jparise:
@scoiatael, just checking in on this. Did you comments make sense?

scoiatael:
Yeah, thanks for them. Simply didn't have time recently to apply them.

Best regards,

Łukasz Czaplinski

pymemcache/client/hash.py
scoiatael: That's one way of viewing it. The other is that this function works regardless of whether we have dead clients or not - meaning the check for empty clients is simply an optimisation.

I'll update the comment accordingly.

jparise:
Thank you, @scoiatae!

PR #276, opened 2020-04-02
Add TLS support for TCP sockets

moisesguimaraes:
This PR is an attempt to add TLS support to pymemcache following the same approach I used to add TLS support to python-binary-memcached.


moisesguimaraes:
@jparise, sure. I just wanted to raise the PR for discussion.

I can see that you don't automatically run memcached in your integration tests and I couldn't find it in the CI. I'd need to pass the key/cert to memcached and use the CA to build the context.

Or you do mean coverage with just mocks?

jparise:
> I can see that you don't automatically run memcached in your integration tests and I couldn't find it in the CI. I'd need to pass the key/cert to memcached and use the CA to build the context.

We _do_ start a memcached service instance for our tests:

https://github.com/jparise/pymemcache/blob/b5d586092fced9e5483fcd03f034c8478376ed72/.github/workflows/ci.yml#L40-L43

... but I'm not sure off the top of my head how to configure it with TLS secrets.

> Or you do mean coverage with just mocks?

This could be the backup testing strategy.

moisesguimaraes:
ah thanks, I was looking for a check called integration in the CI and didn't see that file before.

moisesguimaraes:
@jparise the server params are something like this:

https://github.com/moisesguimaraes/poc-secure-memcached/blob/4081c173f01458f8fe74880ec7cbea5a1481deaa/docker-compose.yaml#L16-L25

but the memcached service must be compiled with --enable-tls. Where does that image comes from? Is it dockerhub latest? If so, I added TLS support to it a couple of months ago and it is fine, otherwise I need another plan.

jparise:
> @jparise the server params are something like this:
> 
> [moisesguimaraes/poc-secure-memcached:docker-compose.yaml@`4081c17`#L16-L25](https://github.com/moisesguimaraes/poc-secure-memcached/blob/4081c173f01458f8fe74880ec7cbea5a1481deaa/docker-compose.yaml#L16-L25)

I believe we can pass options to the containerized service like this:

```yaml
memcached:
  image: memcached:latest
  ports:
  - 11211/tcp
  options: --foo --bar --baz
```
> but the memcached service must be compiled with --enable-tls. Where does that image comes from? Is it dockerhub latest? If so, I added TLS support to it a couple of months ago and it is fine, otherwise I need another plan.

The `image` key specifies the Docker image to use, so I believe `memcachd:latest` means DockerHub latest here.

.github/workflows/ci.yml
```diff
@@ -41,3 +41,14 @@ jobs:
         image: memcached:latest
         ports:
         - 11211/tcp
+      tls_memcached:
+        image: memcached:latest
+        ports:
+        - 11212:11211/tcp
+        volumes:
+        - ./extras/certs/gen:/opt/certs
+        command: >-
```
jparise, line 50: I think you want `options:` instead of `command:`?

moisesguimaraes: I tried with options, it passes them as options to the docker run command, not to the container itself. I'm creating a convenience docker image to have all the needed bits inside as I'm doing this over and over across several projects.

moisesguimaraes:
ok, now the service is in place, I'll start working on the tests.

moisesguimaraes:
@jparise looks like the CI got stuck, how can I ask it to recheck?

Also I still need to investigate TLS using gevent on python2.

jparise:
> @jparise looks like the CI got stuck, how can I ask it to recheck?

I just restarted it. (I have a Re-run option here: https://github.com/pinterest/pymemcache/pull/276/checks)

pymemcache/client/base.py
```diff
@@ -291,6 +293,12 @@ def _connect(self):
             if self.no_delay and sock.family == self.socket_module.AF_INET:
                 sock.setsockopt(self.socket_module.IPPROTO_TCP,
                                 self.socket_module.TCP_NODELAY, 1)
+
+            if self.tls_context is not None and self.socket_module.AF_INET:
+                sock = self.tls_context.wrap_socket(
+                    sock, server_hostname=self.server[0],
+                    do_handshake_on_connect=False,
```
jparise, line 300: I thought `do_handshake_on_connect=False` requires us to call `SSLSocket.do_handshake()` ourselves, but I don't see the code doing that anywhere.

moisesguimaraes: I'll try moving the wrapping before the connection and removing that param. I guess it probably does the handshake in the first write/read to the socket if not connected yet.

jparise: I _think_ we'd want to wrap on `connect`. That's the behavior I'd expect from the connection pools, at least, where we want to pay the cost of establishing the connection upfront rather than on first use.

moisesguimaraes:
Lets enable back the py2 TLS tests and see what happens there.

.github/workflows/ci.yml
```diff
@@ -34,10 +34,17 @@ jobs:
...
 
     services:
       memcached:
         image: memcached:latest
         ports:
         - 11211/tcp
+      tls_memcached:
+        image: scoriacorp/tls_memcached:latest
+        ports:
+        - 11211/tcp
```
jparise, line 50: Do you want to run this one `11212`, which is the default you have listed for the tests?

moisesguimaraes: 11211 is inside the container. When I'm testing locally I usually do

```
docker run --rm -p 11212:11211 scoriacorp/tls_memcached
```

cause the non-TLS memcache will be already running on port 11211.

moisesguimaraes:
Nice, all tests passed now for all versions. I'll work on the docs on Monday o/

moisesguimaraes:
@jparise, if merged, do you have a timeline for the next release?

jparise:
Thanks again, @moisesguimaraes!

I think we can release this pretty quickly. I put together a [3.1 milestone](https://github.com/pinterest/pymemcache/milestone/3) to track things that will be part of that release.

jparise:
@moisesguimaraes version 3.1 has been tagged and released.

PR #279, opened 2020-04-24
Align Client and MockMemcacheClient to ease third-party testing #279

PaulRenvoise:
This PR makes `MockMemcacheClient` closer to the implementation of `Client`, namely:

- Added `socket_module`, and `tls_context` parameters `__init__()`
- Added `socket_module`, `sock`, and `tls_context` attributes
- Added `replace()`
- Added `cas()` (which raises `MemcacheClientError` as `cas_enabled` is false for this `MockMemcacheClient`)
- Added `touch()`
- Added `cache_memlimit()`
- Added `version()`
- Added `flush_all()` (internally calling `clear()`)
- Added `quit()`
- Added `close()`
- Refactored `set()` to have default `flags` set to None
- Refactored `set_many()` to follow `Client`'s implementation (always return `[]` when `noreply` is True, else return the keys that were not inserted)
- Refactored `decr()` to have a similar implementation to `incr()` (also taking in account `noreply`'s value)
- Refactored `add()` to take in account `noreply`'s value
- Refactored `get_many()` to support `expire` and `flags`
- Refactored `stats()` to accepts arguments
- Refactored `append()`, `prepend()`, `set_many()`, and `add()` to forward `expire`, `noreply`, and `flags` to their internal call to `set()`
- Set all default value of `noreply` to True when it was None (since in Client, it uses `default_noreply` (set to True), when `noreply` is None)
- Set all default value of `expire` to 0

+ Bumped version to 3.1.1

### DIFF
--- a/pymemcache/__init__.py
+++ b/pymemcache/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.1.1'
+__version__ = '3.1.0'
 
 from pymemcache.client.base import Client  # noqa
 from pymemcache.client.base import PooledClient  # noqa
--- a/pymemcache/__init__.py
+++ b/pymemcache/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.1.0'
+__version__ = '3.1.1'
 
 from pymemcache.client.base import Client  # noqa
 from pymemcache.client.base import PooledClient  # noqa
--- a/pymemcache/test/utils.py
+++ b/pymemcache/test/utils.py
@@ -8,8 +8,9 @@ This module is considered public API.
 import time
 
 import six
+import socket
 
-from pymemcache.exceptions import MemcacheIllegalInputError
+from pymemcache.exceptions import MemcacheIllegalInputError, MemcacheClientError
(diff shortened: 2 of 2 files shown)


pymemcache/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '3.1.0'
+__version__ = '3.1.1'
```
jparise, line 1: Please remove this part of the change. We'll bump the version as part of the release process.

pymemcache/test/utils.py
```diff
@@ -8,8 +8,9 @@
 import time
 
 import six
+import socket
 
-from pymemcache.exceptions import MemcacheIllegalInputError
+from pymemcache.exceptions import MemcacheIllegalInputError, MemcacheClientError
```
jparise, line 13: Minor, but let's list `MemcacheClientError` first (alphabetically).

PaulRenvoise:
Thanks for the quick response, everything should be good now!

jparise:
Thanks for the contribution! I'll try to get this out in a patch release soon.

PR #280, opened 2020-04-28
Make PooledClient and HashClient able to use a custom client class

bloodearnest:
The user can subclass Client (e.g to add telemetry or logging), and have
their custom client class used by PooledClient and HashClient to create
new clients.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -1009,6 +1009,9 @@ class PooledClient(object):
     in the pool. Your serde object must therefore be thread-safe.
     """
 
+    #: :class:`Client` class used to create new clients
+    client_class = Client
+
     def __init__(self,
                  server,
                  serde=None,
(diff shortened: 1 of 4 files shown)


pymemcache/client/base.py
```diff
@@ -1054,19 +1057,20 @@ def check_key(self, key):
...
-                        no_delay=self.no_delay,
-                        # We need to know when it fails *always* so that we
-                        # can remove/destroy it from the pool...
-                        ignore_exc=False,
-                        socket_module=self.socket_module,
-                        key_prefix=self.key_prefix,
-                        default_noreply=self.default_noreply,
-                        allow_unicode_keys=self.allow_unicode_keys,
-                        tls_context=self.tls_context)
+        client = self.client_class(
```
jparise, line 1060: While we're here, I think this could just `return` the object directly.

pymemcache/client/base.py
```diff
@@ -1009,6 +1009,9 @@ class PooledClient(object):
     in the pool. Your serde object must therefore be thread-safe.
     """
 
+    # default class to use for creating clients
```
jparise, line 1012: I _think_ we want to use Sphinx's `#:` comment syntax so this shows up in the generated API documentation. Also, `Client` is the default class, but the attribute holds the class that's actually going to be used.

```suggestion
    #: :class:`Client` class used to create new clients
```

pymemcache/client/hash.py
```diff
@@ -14,6 +14,8 @@ class HashClient(object):
     """
     A client for communicating with a cluster of memcached servers
     """
+    # default class to use for clients
```
jparise, line 17:
```suggestion
    #: :class:`Client` class used to create new clients
```

pymemcache/client/base.py
bloodearnest: :+1:

bloodearnest:
Would you like me to squash these changes? Not sure what your merge policy is! :)

jparise:
> Would you like me to squash these changes? Not sure what your merge policy is! :)

Yes, please. I'll merge afterward.

bloodearnest:
Squashed and pushed. Thanks for the quick response. Do you plan a release soon?

jparise:
> Squashed and pushed. Thanks for the quick response. Do you plan a release soon?

I just released 3.1.1 today, which only had internal test-level changes. I'll hold your change for a 3.2.0 release because it introduces new public API.

PR #295, opened 2020-09-01
Added support for UNIX sockets in HashClient.

pope1ni:
`HashClient` didn't support using paths to UNIX sockets and passing them to the underlying client.
With these changes it should also be possible to use a mix of UNIX sockets and `(host, port)` 2-tuples.

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -46,7 +46,8 @@ class HashClient(object):
         Constructor.
 
         Args:
-          servers: list(tuple(hostname, port))
+          servers: list() of tuple(hostname, port) or string containing a UNIX
+                   socket path.
           hasher: optional class three functions ``get_node``, ``add_node``,
                   and ``remove_node``
                   defaults to Rendezvous (HRW) hash.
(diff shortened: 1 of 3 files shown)


pope1ni:
Yes, I thought about doing this. Would you also like a `DeprecationWarning` to be triggered if `port` is set?

jparise:
> Yes, I thought about doing this. Would you also like a `DeprecationWarning` to be triggered if `port` is set?

`port` would still be used for TCP-based connections, as it is today, so there isn't a deprecation case that we need to announce.

pope1ni:
Ok. I've restored the original signature to those methods and added an extra test.

pymemcache/test/test_client_hash.py
```diff
@@ -354,7 +355,27 @@ class MyClient(Client):
...
+        client.add_server(('host', 11211))
         assert isinstance(client.clients['host:11211'], MyClient)
 
+    def test_mixed_inet_and_unix_sockets(self):
+        servers = [
+            '/tmp/pymemcache.{pid}'.format(pid=os.getpid()),
+            ('127.0.0.1', 11211),
+        ]
+        client = HashClient(servers)
+        assert set(servers) == set(c.server for c in client.clients.values())
```
jparise, line 367: You can simplify the right-hand side using a set comprehension:

```suggestion
        assert set(servers) == {c.server for c in client.clients.values()}
```

pymemcache/client/hash.py
```diff
@@ -101,30 +102,41 @@ def __init__(
...
         self.encoding = encoding
         self.tls_context = tls_context
 
-    def add_server(self, server, port):
-        key = '%s:%s' % (server, port)
+    def add_server(self, server, port=None):
+        # To maintain backward compatibility, if a port is provided, assume
+        # that server wasn't provided as a (host, port) tuple.
+        if port is not None:
+            server = (server, port)
```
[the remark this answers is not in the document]
pope1ni: Sure. I've also updated to use `six.text_type` as I note that we're still supporting Python 2.

pymemcache/test/test_client_hash.py
pope1ni: :man_facepalming:

pope1ni:
Ok. I've made those changes. While I was at it I added a small cleanup for `isinstance()` checks noticed a bug in support for custom client classes when `use_pooling=True` for `HashClient` which I've also fixed.

pymemcache/test/test_client_hash.py
```diff
@@ -42,7 +43,7 @@ def test_setup_client_without_pooling(self):
         client_class = 'pymemcache.client.hash.HashClient.client_class'
         with mock.patch(client_class) as internal_client:
             client = HashClient([], timeout=999, key_prefix='foo_bar_baz')
-            client.add_server('127.0.0.1', '11211')
+            client.add_server(('127.0.0.1', '11211'))
```
jparise, line 46: It doesn't look like we have any tests that use the `.add_server(server, port=port)` signature now?

pope1ni: https://github.com/pinterest/pymemcache/pull/295/files#diff-dfcf37102448920749549449cc837e8aR382-R399

jparise: Perfect. Thanks for the link!

jparise (approved):

pope1ni:
Hi @jparise. Would it be possible to make a release so that this fix is available? Thanks.

jparise:
@pope1ni yes, I'll run through the release process this morning.

pope1ni:
@jparise Thank you!

PR #300, opened 2020-10-11
Drop Python version 3.5 and add 3.9

jparise:
- https://pythoninsider.blogspot.com/2020/10/python-35-is-no-longer-supported.html
- https://pythoninsider.blogspot.com/2020/10/python-390-is-now-available-and-you-can.html

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,7 +1,7 @@
+future
 mock
 pytest
 pytest-cov
-gevent==1.3.6; "PyPy" not in platform_python_implementation
+gevent==20.9.0; "PyPy" not in platform_python_implementation
 pylibmc; sys.platform != 'win32'
 python-memcached
-future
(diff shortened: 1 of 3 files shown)


test-requirements.txt
```diff
@@ -1,7 +1,7 @@
+future
 mock
 pytest
 pytest-cov
-gevent==1.3.6; "PyPy" not in platform_python_implementation
+gevent==20.9.0; "PyPy" not in platform_python_implementation
 pylibmc; sys.platform != 'win32'
 python-memcached
-future
```
jogo, line 7: https://github.com/pinterest/pymemcache/search?q=builtins isn't this still used?

jparise: It is. I just sorted it to the top of the list (alphabetically).

jogo: doh

jogo (approved):

PR #309, opened 2021-03-04
Remove idle connections from pool

jwgcarlson:
The Memcached server can be configured to drop idle connections (via the
`idle_timeout` option). When this occurs, pymemcache may still try to
use the connection, resulting in `MemcacheUnexpectedCloseError`. Hence
the need for client-side idle timeout logic.

This change was proposed in pinterest/pymemcache#114, but ultimately dropped.
The current PR uses the same approach but differs slightly in the details.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -1039,8 +1039,9 @@ class PooledClient(object):
       max_pool_size: maximum pool size to use (going above this amount
                      triggers a runtime error), by default this is 2147483648L
                      when not provided (or none).
-      pool_idle_timeout: pooled connections are dropped if unused for this
-                         amount of time (default None)
+      pool_idle_timeout: pooled connections are discarded if they have been
+                         unused for this many seconds. A value of 0 or None
+                         indicates that pooled connections are never discarded.
       lock_generator: a callback/type that takes no arguments that will
                       be called to create a lock or semaphore that can
                       protect the pool from concurrent access (for example a
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -1039,6 +1039,8 @@ class PooledClient(object):
       max_pool_size: maximum pool size to use (going above this amount
                      triggers a runtime error), by default this is 2147483648L
                      when not provided (or none).
+      pool_idle_timeout: pooled connections are dropped if unused for this
+                         amount of time (default None)
       lock_generator: a callback/type that takes no arguments that will
                       be called to create a lock or semaphore that can
                       protect the pool from concurrent access (for example a
@@ -1065,6 +1067,7 @@ class PooledClient(object):
                  socket_module=socket,
                  key_prefix=b'',
                  max_pool_size=None,
+                 pool_idle_timeout=None,
                  lock_generator=None,
                  default_noreply=True,
                  allow_unicode_keys=False,
@@ -1088,6 +1091,7 @@ class PooledClient(object):
             self._create_client,
             after_remove=lambda client: client.close(),
             max_size=max_pool_size,
+            idle_timeout=pool_idle_timeout,
             lock_generator=lock_generator)
         self.encoding = encoding
         self.tls_context = tls_context
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -1040,8 +1040,8 @@ class PooledClient(object):
                      triggers a runtime error), by default this is 2147483648L
                      when not provided (or none).
       pool_idle_timeout: pooled connections are discarded if they have been
-                         unused for this many seconds. A value of 0 or None
-                         indicates that pooled connections are never discarded.
+                         unused for this many seconds. A value of 0 indicates
+                         that pooled connections are never discarded.
       lock_generator: a callback/type that takes no arguments that will
                       be called to create a lock or semaphore that can
                       protect the pool from concurrent access (for example a
@@ -1068,7 +1068,7 @@ class PooledClient(object):
                  socket_module=socket,
                  key_prefix=b'',
                  max_pool_size=None,
-                 pool_idle_timeout=None,
+                 pool_idle_timeout=0,
                  lock_generator=None,
                  default_noreply=True,
                  allow_unicode_keys=False,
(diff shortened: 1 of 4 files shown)


pymemcache/pool.py
```diff
@@ -63,19 +67,26 @@ def get_and_release(self, destroy_on_fail=False):
 
     def get(self):
         with self._lock:
-            if not self._free_objs:
+            # Find a free object, removing any that have idled for too long.
+            now = self._idle_clock()
+            while self._free_objs:
+                obj = self._free_objs.popleft()
+                if now - obj._last_used > self.idle_timeout:
+                    self._after_remove(obj)
```
cgordon, line 75: _after_remove can be None.

pymemcache/pool.py
```diff
@@ -63,19 +67,26 @@ def get_and_release(self, destroy_on_fail=False):
...
+                else:
+                    break
```
cgordon, line 77: Aesthetically, I would prefer the loop to be:

```
if now - obj._last_used <= self.idle_timeout:
    break

if self._after_remove is not None:
    self._after_remove(obj)
```

cgordon (changes requested):
I'm fine with the added functionality, but I want to note the following caveats:

- Users can still get `MemcachedUnexpectedCloseError`, even with this change. This is a distributed system, so it will be very easy for the timing to work out such that the connection looks "live" on the client side, but has become "idle" on the server side.
- Users still need to expect and handle `MemcacheUnexpectedCloseError`. I'm starting to think, after recent discussions, that we need to provide more guidance for how to do this. There is no way to guarantee that you will never get this exception.

With those caveats, I think this can eliminate some spurious cases of `MemcacheUnexpectedCloseError`, which should make normal error handling a little more stream-lined.

jwgcarlson:
Thank you for your prompt review, and I apologize for my own delayed response. I agree with all your comments, and I'll work on updating the PR as soon as I find the time.

jparise:
> Thank you for your prompt review, and I apologize for my own delayed response. I agree with all your comments, and I'll work on updating the PR as soon as I find the time.

Just wanted to check in and see if you were able to find the time to make these revisions.

pymemcache/client/base.py
```diff
@@ -1036,6 +1036,8 @@ class PooledClient(object):
       max_pool_size: maximum pool size to use (going above this amount
                      triggers a runtime error), by default this is 2147483648L
                      when not provided (or none).
+      pool_idle_timeout: pooled connections are dropped if unused for this
```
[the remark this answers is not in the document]
jwgcarlson: Done.

pymemcache/pool.py
jwgcarlson: Done.

jwgcarlson:
> Just wanted to check in and see if you were able to find the time to make these revisions.

Oops, I definitely forgot about it, but made the requested changes now. Thanks for your patience.

pymemcache/client/base.py
```diff
@@ -1039,6 +1039,9 @@ class PooledClient(object):
       max_pool_size: maximum pool size to use (going above this amount
                      triggers a runtime error), by default this is 2147483648L
                      when not provided (or none).
+      pool_idle_timeout: pooled connections are discarded if they have been
+                         unused for this many seconds. A value of 0 or None
+                         indicates that pooled connections are never discarded.
```
jparise, line 1044: Given that `0` is as safe of a "truthy" sentinel value as `None` (and we upgrade `None` to `0` internally), I think we should always say this value is an `int` that defaults to `0`. What do you think?

jwgcarlson: Sure, I'm fine with that. Should `max_pool_size` be handled the same way?

jparise: Probably, but we've already shipped `max_pool_size` in public releases, so it's best to leave it be to avoid accidentally breaking existing installations.

pymemcache/client/hash.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         socket_module=socket,
         key_prefix=b'',
         max_pool_size=None,
+        pool_idle_timeout=None,
```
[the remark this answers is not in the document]
jwgcarlson: Done.

pymemcache/test/test_client.py
```diff
@@ -1255,6 +1255,35 @@ class MyClient(Client):
...
+    def make_client(self, mock_socket_values, **kwargs):
+        mock_client = Client(None, **kwargs)
+        mock_client.sock = MockSocket(list(mock_socket_values))
+        client = PooledClient(None, pool_idle_timeout=60, **kwargs)
+        client.client_pool = pool.ObjectPool(lambda: mock_client)
+        return client
+
+    def test_free_idle(self):
+        num_removed = 0
+        def increment_removed(obj):
```
jparise, line 1268: The CI checks are failing here due to this flake8 formatting rule:

```
./pymemcache/test/test_client.py:1268:9: E306 expected 1 blank line before a nested definition, found 0
        def increment_removed(obj):
```

jwgcarlson: Oops, fixed.

pymemcache/test/test_client.py
```diff
@@ -1255,6 +1255,36 @@ class MyClient(Client):
...
+        mock_client.sock = MockSocket(list(mock_socket_values))
+        client = PooledClient(None, pool_idle_timeout=60, **kwargs)
+        client.client_pool = pool.ObjectPool(lambda: mock_client)
+        return client
+
+    def test_free_idle(self):
+        num_removed = 0
+
+        def increment_removed(obj):
+            nonlocal num_removed
```
jparise, line 1270: `nonlocal` isn't supported under Python 2.x.

jparise (approved):
Looks good to me!

@cgordon any remaining concerns from your end?

pymemcache/test/test_client.py
jwgcarlson: Okay, changed to use a `Counter` object instead.

cgordon (approved):
LGTM!

jparise:
Thanks @jwgcarlson!

PR #311, opened 2021-04-13
Removing trailing space after a command if there are no arguments.

jerrymiu:
According to the [Memcached Protocol](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L1151), the no argument form of the stats command should have no trailing spaces. Currently, `b'stats \r\n'` instead of `b'stats\r\n'` is being sent to the socket which leads to a malformed response.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -893,8 +893,10 @@ class Client(object):
         remapped_keys = dict(zip(prefixed_keys, keys))
 
         # It is important for all keys to be listed in their original order.
-        # For Memcached Protocol correctness, if the list of prefixed_keys is empty, the name of the command will be immediately followed by the end of line.
-        cmd = name + b' ' + b' '.join(prefixed_keys) + b'\r\n' if prefixed_keys else name + b'\r\n'
+        cmd = name
+        if prefixed_keys:
+            cmd += b' ' + b' '.join(prefixed_keys)
+        cmd += b'\r\n'
 
         try:
             if self.sock is None:
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -893,7 +893,8 @@ class Client(object):
         remapped_keys = dict(zip(prefixed_keys, keys))
 
         # It is important for all keys to be listed in their original order.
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -893,7 +893,8 @@ def _fetch_cmd(self, name, keys, expect_cas):
         remapped_keys = dict(zip(prefixed_keys, keys))
 
         # It is important for all keys to be listed in their original order.
-        cmd = name + b' ' + b' '.join(prefixed_keys) + b'\r\n'
+        # For Memcached Protocol correctness, if the list of prefixed_keys is empty, the name of the command will be immediately followed by the end of line.
+        cmd = name + b' ' + b' '.join(prefixed_keys) + b'\r\n' if prefixed_keys else name + b'\r\n'
```
jparise, line 897: This is pretty complex. How about:

```py
cmd = name
if prefixed_keys:
    cmd += b' ' + b' '.join(prefixed_keys)
cmd += b'\r\n'
```

Then I also don't think you need a comment like about "protocol correctness" because the condition is obvious in the code.

jerrymiu: Seems simpler, sounds good thanks!

jparise:
It looks like you'll also need to update the `stats` tests with the new (correct) behavior:

```
pymemcache/test/test_client.py:925: in test_stats_conversions
    assert client.sock.send_bufs == [
E   AssertionError: assert [b'stats\r\n'] == [b'stats \r\n']
E     At index 0 diff: b'stats\r\n' != b'stats \r\n'
E     Full diff:
E     - [b'stats \r\n']
E     ?         -
E     + [b'stats\r\n']
```

jerrymiu:
Noted, thanks I've updated the test cases and I got the automated checks to pass.

jparise (approved):

jparise:
> Noted, thanks I've updated the test cases and I got the automated checks to pass.

Good stuff. Thanks for your contribution!

jerrymiu:
You're welcome!

PR #315, opened 2021-05-07
Fixed `HashClient.{get,set}_many()` with UNIX sockets.

### LINKED ISSUE #314: TypeError when calling 'get_many' on HashClient with Unix socket location
```python
from pymemcache.client.hash import HashClient

client = HashClient(['/path/to/memcached.socket', ])
client.get_many(['ITEM_A', 'ITEM_B'])

  File "<input>", line 1, in <module>
  File "/path/to/venv/lib/python3.9/site-packages/pymemcache/client/hash.py", line 413, in get_many
    if gets:
TypeError: not enough arguments for format string
```

At line 409 of [/pymemcache/client/hash.py](https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/hash.py#L409)
```python
client = self.clients['%s:%s' % server]
```
Since `server` is  a `str` of Unix socket location in this case, not the `tuple` that contains `IP` and `PORT`, `TypeError: not enough arguments for format string` happens when using `get_many` and `set_many`.

Should be like
```python
if isinstance(server, six.string_types):
    client = self.clients[server]
else:
    client = self.clients['%s:%s' % server]
```

or restricting Unix Socket on `HashClient` if it's not likely to have multiple Unix sockets (I'm not sure about this)

pope1ni:
Overlooked when UNIX socket support was added to `HashClient` in
acd962b586a4fe018a00acb7e1621be373c13c3b.

Fixes #314.

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -112,6 +112,11 @@ class HashClient(object):
         self.encoding = encoding
         self.tls_context = tls_context
 
+    def _make_client_key(self, server):
+        if isinstance(server, (list, tuple)) and len(server) == 2:
+            return '%s:%s' % server
+        return server
+
     def add_server(self, server, port=None):
         # To maintain backward compatibility, if a port is provided, assume
         # that server wasn't provided as a (host, port) tuple.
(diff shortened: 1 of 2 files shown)


pymemcache/client/hash.py
```diff
@@ -383,7 +382,10 @@ def set_many(self, values, *args, **kwargs):
             client_batches[client.server][key] = value
 
         for server, values in client_batches.items():
-            client = self.clients['%s:%s' % server]
+            if isinstance(server, six.string_types):
+                client = self.clients[server]
+            else:
+                client = self.clients['%s:%s' % server]
```
jparise, line 388: I wonder whether these conditions would be slightly better if they were inverted and used `isinstance(server, tuple)` instead. That's the stronger precondition/invariant given the `%` formatting performed in that branch. What do you think?

jparise (approved):
Looks great - thanks!

PR #320, opened 2021-05-27
Use client.set_many in HashClient.

### LINKED ISSUE #319: `HashClient.set_many` performs several single `set` commands
This is not strictly an issue, more of a clarification before I fork something and start working.

The current `HashClient.set_many()` function ends up performing `n` single `set` commands for `n` values, instead of performing `m` `set_many` commands for `m` number of `base.Client` in the `HashClient`. Assuming we run without the `PooledClient`.

The call chain goes `HashClient.set_many -> HashClient._safely_run_set_many -> HashClient._set_many -> base.Client.set`
My expectations would be that the end should be `base.Client.set_many` instead. The line in question: [L340](https://github.com/pinterest/pymemcache/blob/64109c2b140807d19cb65ce8f34ff0121b468fe3/pymemcache/client/hash.py#L340)

I was going to take a stab at implementing it that way but just wanted to hear if this had already been tried and scrapped, and if there was a specific reason for using sequential `set` commands instead of `set_many`?

martinnj:
This should fix #319 .

Some unit tests had to be changed since the data returned from `client.set_many` is different from when we iterate over `client.set` multiple times.

- `test_ignore_exec_set_many` - With the old approach if an exception occurred we could mark the individual key as failed and keep going. With `set_many` we only get a list of failed keys if no exception was thrown. Instead if `ignore_exc` is set, I chose to mark all keys as set.
- `test_set_many_unix` - It relies on the ordering of keys in a dict, which is unordered. I changed it from checking if a specific key failed, to just check a single key failed.

I don't have all the python interpreters installed on my machine, but tests pass for the following envs in tox:

- `py27`
- `py36`
- `py37`
- `py27-flake8`
- `py39-flake8`

`py39` tests failed because it was compiled without sqlite.

If I missed something or there is anything else that needs doing, please let me know.

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -336,15 +336,12 @@ class HashClient(object):
         succeeded = []
 
         try:
-            for key, value in six.iteritems(values):
-                result = client.set(key, value, *args, **kwargs)
-                if result:
-                    succeeded.append(key)
-                else:
-                    failed.append(key)
+            failed = client.set_many(values, *args, **kwargs)
         except Exception as e:
-            return succeeded, failed, e
+            if not self.ignore_exc:
+                return succeeded, failed, e
 
+        succeeded = [key for key in six.iterkeys(values) if key not in failed]
         return succeeded, failed, None
 
     def close(self):
--- a/pymemcache/test/test_client_hash.py
+++ b/pymemcache/test/test_client_hash.py
@@ -301,7 +301,7 @@ class TestHashClient(ClientTestMixin, unittest.TestCase):
         ], ignore_exc=True)
         result = client.set_many(values, noreply=False)
 
-        assert len(result) == 2
+        assert len(result) == 0
 
     def test_noreply_set_many(self):
         values = {
@@ -336,7 +336,7 @@ class TestHashClient(ClientTestMixin, unittest.TestCase):
         ])
 
         result = client.set_many(values, noreply=False)
-        assert result == ['key2']
+        assert len(result) == 1
 
     def test_server_encoding_pooled(self):
         """


jparise:
> `test_set_many_unix` - It relies on the ordering of keys in a dict, which is unordered. I changed it from checking if a specific key failed, to just check a single key failed.

I think this would be a good place to use [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict). Otherwise, the test isn't particularly strong because it doesn't verify that the `NOT_STORED` item is the one that's actually returned.

martinnj:
> 
> 
> > `test_set_many_unix` - It relies on the ordering of keys in a dict, which is unordered. I changed it from checking if a specific key failed, to just check a single key failed.
> 
> I think this would be a good place to use [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict). Otherwise, the test isn't particularly strong because it doesn't verify that the `NOT_STORED` item is the one that's actually returned.

I took a look at this, it would require the [Line 369](https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/hash.py#L369) to be changed to `collections.defaultdict(collections.OrderedDict)`.
In this case the issue is caused by the `dict` class only being ordered from 3.6 and up (guaranteed from 3.7).
By switching to `OrderedDict` all versions will slow down, in order to satisfy a failing test for `py27, py34 and py35`.

I changed the benchmark to use the `HashClient` instead of the regular one, and got the following results from this command:
```sh
pytest --verbose --capture=no --no-cov -m benchmark pymemcache/test --keys 100 --count 10000 -k set_multi
```
- With `dict`: 41.0434 seconds.
- With `OrderedDict`: 41.7659 seconds.

Let me know what you think.

jparise:
> > > `test_set_many_unix` - It relies on the ordering of keys in a dict, which is unordered. I changed it from checking if a specific key failed, to just check a single key failed.
> > 
> > 
> > I think this would be a good place to use [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict). Otherwise, the test isn't particularly strong because it doesn't verify that the `NOT_STORED` item is the one that's actually returned.
> 
> I took a look at this, it would require the [Line 369](https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/hash.py?rgh-link-date=2021-05-27T18%3A45%3A40Z#L369) to be changed to `collections.defaultdict(collections.OrderedDict)`.
> In this case the issue is caused by the `dict` class only being ordered from 3.6 and up (guaranteed from 3.7).
> By switching to `OrderedDict` all versions will slow down, in order to satisfy a failing test for `py27, py34 and py35`.

That makes sense. I originally thought the issue was only with the dictionary provided by the test function.

Thanks for the additional investigation!

jparise (approved):

martinnj:
@jparise - Sorry to nag a bit, is there any chance this can go into a small 3.4.4 release before the 4.0 milestone I see in the PR tab?
Have some legacy 2.7 code I'd like to have the change until it's ready to go full Python 3. :)

jparise:
> @jparise - Sorry to nag a bit, is there any chance this can go into a small 3.4.4 release before the 4.0 milestone I see in the PR tab?
> Have some legacy 2.7 code I'd like to have the change until it's ready to go full Python 3. :)

Yes, I'll publish a release shortly (#322).

Issue #307, opened 2021-02-11
Implementing a reconnection logic after MemcacheUnexpectedCloseError


cayolblake:
In the occasion of a memcached server restart, on the first access attempt after server restarted, we get "MemcacheUnexpectedCloseError" and the request fails, then the next following access (a totally different request) goes through properly.

What am I missing here? Do I need to implement some reconnection logic on failure? Any hints on doing so if that's the case?

cayolblake:
Well, after a connection breaks, contrary to what you stated, it doesn't attempt to reconnect on its own, and that's my main issue.

cgordon:
> In the occasion of a memcached server restart, on the first access attempt after server restarted, we get "MemcacheUnexpectedCloseError" and the request fails, then the next following access (a totally different request) goes through properly.

As you noted, "the next following access [..] goes through properly". That is pymemcache re-establishing the connection automatically for you.

cayolblake:
I'll try to clarify a bit more.

1. Operation is normal.
2. Memcached server restarts.
3. First operation after restart fails with "MemcacheUnexpectedCloseError". It fails.
4. Second operation goes fine.

My expectation is on step 3 it reconnects and let the operation go through taking into consideration the overhead of the reconnection.

cgordon:
I will also try to clarify more.

There are many ways to use memcached, and there is no one-size-fits-all error handling solution that will work for everyone. Because of that, pymemcache reports errors to users, and expects users to handle those errors appropriately for their use case. As a convenience, if you continue to make requests through the pymemcache Client, it will attempt to reconnect a closed socket for you. That just saves you the trouble of needing to create a new Client instance, or call an explicit "reconnect" API. The automatic retries are not designed to paper over issues connecting to memcached. I believe that explicit errors are the best way for this API to communicate with its users when an issue occurs.

In my opinion, it would be a bad idea for pymemcache to add automatic retries into the core Client library. I strongly prefer that users of the library see exceptions rather than high latency (as would happen with a retry/backoff loop). I tried to outline some of the possible ways that errors can be handled in my long post above. The error case you mentioned is a good candidate for automatic retries, but it is not the only kind of error experienced by memcached users (and it is very unlikely to be the only kind of error you experience, particularly if your request rate increases over time).

If you want automatic retries then it would be relatively easy to add a configurable Client wrapper that automatically retried requests. We would be happy to review that as a submission if you are so inclined.

4383:
Hello,

[...]

PR #321, opened 2021-05-27
Drop python support for 2.7, 3.4 and 3.5

jogo:
End of Life information:
* 3.5 https://www.python.org/downloads/release/python-350/
* 3.4 https://www.python.org/downloads/release/python-340/
* 2.7 https://www.python.org/doc/sunset-python-2/

Remove six dependency and run pyupgrade to start using native python 3 syntax

PR #324, opened 2021-06-17
Implement RetryingClient

martinnj:
Goal here is basically to help with https://github.com/pinterest/pymemcache/issues/307.
And takes inspiration from this [comment](https://github.com/pinterest/pymemcache/issues/307#issuecomment-862109935)

The idea is a thin "wrapper client" that does nothing but forward calls and retry as appropriate.
The features are as follows:
 - Takes any Client or Client-like class.
 - Configurable number of retries.
 - Configurable sleep between retries. (Don't think it's really a good idea in a cache, but it was easy, so here we are.)
 - Allows retrying only for specific Exception types.
 - Allows skipping retries for specific Exception types.

I haven't written tests yet since I wanted some opinions on the interface/constructor first.

Currently the class uses `__getattr__` to pass through any attributes/methods that isn't present on itself to the inner client.
It also passes through `dir()` calls to the inner client to allow for checking available methods, attributes, etc.
Is this an acceptable implementation or is it too opaque?

I also did a more naive implementation, it is commented out in the code.

Feedback and suggestions are welcome.
I'll write the unit-tests ones the feel of the client is more solid.

@cgordon are you the person to go to for this? 😄

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -77,6 +77,34 @@ on if a server goes down.
     client.set('some_key', 'some value')
     result = client.get('some_key')
 
+Using the built-in retrying mechanism
+-------------------------------------
+The library comes with retry mechanisms that can be used to wrap all kind of
+pymemcache clients. The wrapper allow you to define the exceptions that you want
+to handle with retries, which exceptions to exclude, how many attempts to make
+and how long to wait between attemots.
+
+The ``RetryingClient`` wraps around any of the other included clients and will
+have the same methods. For this example we're just using the base ``Client``.
+
+.. code-block:: python
+
+    from pymemcache.client.base import Client
+    from pymemcache.client.retrying import RetryingClient
+    from pymemcache.exceptions import MemcacheUnexpectedCloseError
+
+    base_client = Client(("localhost", 11211))
+    client = RetryingClient(
+        base_client,
+        attempts=3,
+        retry_delay=0.01,
+        retry_for=[MemcacheUnexpectedCloseError]
+    )
+    client.set('some_key', 'some value')
+    result = client.get('some_key')
+
+The above client will attempt each call three times with a wait of 10ms between
+each attempt, as long as the exception is a ``MemcacheUnexpectedCloseError``.
 
 Using TLS
 ---------
--- a/pymemcache/client/retrying.py
+++ b/pymemcache/client/retrying.py
@@ -1,0 +1,185 @@
+""" Module containing the RetryingClient wrapper class. """
+
+from time import sleep
+
+
+def _ensure_tuple_argument(argument_name, argument_value):
+    """
+    Helper function to ensure the given arguments are tuples of Exceptions (or
+    subclasses), or can at least be converted to such.
+
+    Args:
+      argument_name: str, name of the argument we're checking, only used for
+        raising meaningful exceptions.
+      argument: any, the argument itself.
+
+    Returns:
+      tuple[Exception]: A tuple with the elements from the argument if they are
+        valid.
+
+    Exceptions:
+      ValueError: If the argument was not None, tuple or Iterable.
+      ValueError: If any of the elements of the argument is not a subclass of
+        Exception.
+    """
+
+    # Ensure the argument is a tuple, set or list.
+    if argument_value is None:
+        return tuple()
+    elif not isinstance(argument_value, (tuple, set, list)):
+        raise ValueError(
+            "%s must be either a tuple, a set or a list." % argument_name
+        )
+
+    # Convert the argument before checking contents.
+    argument_tuple = tuple(argument_value)
+
+    # Check that all the elements are actually inherited from Exception.
+    # (Catchable)
+    if not all([issubclass(arg, Exception) for arg in argument_tuple]):
+        raise ValueError(
+            "%s is only allowed to contain elements that are subclasses of "
+            "Exception." % argument_name
+        )
+
+    return argument_tuple
+
+
+class RetryingClient(object):
+    """
+    Client that allows retrying calls for the other clients.
+    """
+
+    def __init__(
+        self,
+        client,
+        attempts=2,
+        retry_delay=0,
+        retry_for=None,
+        do_not_retry_for=None
+    ):
+        """
+        Constructor for RetryingClient.
+
+        Args:
+          client: Client|PooledClient|HashClient, inner client to use for
+            performing actual work.
+          attempts: optional int, how many times to attempt an action before
+            failing. Must be 1 or above. Defaults to 2.
+          retry_delay: optional int|float, how many seconds to sleep between
+            each attempt.
+            Defaults to 0.
+
+          retry_for: optional None|tuple|set|list, what exceptions to
+            allow retries for. Will allow retries for all exceptions if None.
+            Example:
+                `(MemcacheClientError, MemcacheUnexpectedCloseError)`
+            Accepts any class that is a subclass of Exception.
+            Defaults to None.
+
+          do_not_retry_for: optional None|tuple|set|list, what
+            exceptions should be retried. Will not block retries for any
+            Exception if None.
+            Example:
+                `(IOError, MemcacheIllegalInputError)`
+            Accepts any class that is a subclass of Exception.
+            Defaults to None.
+
+        Exceptions:
+          ValueError: If `attempts` is not 1 or above.
+          ValueError: If `retry_for` or `do_not_retry_for` is not None, tuple or
+            Iterable.
+          ValueError: If any of the elements of `retry_for` or
+            `do_not_retry_for` is not a subclass of Exception.
+          ValueError: If there is any overlap between `retry_for` and
+            `do_not_retry_for`.
+        """
+
+        if attempts < 1:
+            raise ValueError(
+                "`attempts` argument must be at least 1. "
+                "Otherwise no attempts are made."
+            )
+
+        self._client = client
+        self._attempts = attempts
+        self._retry_delay = retry_delay
+        self._retry_for = _ensure_tuple_argument("retry_for", retry_for)
+        self._do_not_retry_for = _ensure_tuple_argument(
+            "do_not_retry_for", do_not_retry_for
+        )
+
+        # Verify no overlap in the go/no-go exception collections.
+        for exc_class in self._retry_for:
+            if exc_class in self._do_not_retry_for:
+                raise ValueError(
+                    "Exception class \"%s\" was present in both `retry_for` "
+                    "and `do_not_retry_for`. Any exception class is only "
+                    "allowed in a single argument." % repr(exc_class)
+                )
+
+        # Take dir from the client to speed up future checks.
+        self._client_dir = dir(self._client)
+
+    def _retry(self, name, func, *args, **kwargs):
+        """
+        Workhorse function, handles retry logic.
+
+        Args:
+          name: str, Name of the function called.
+          func: callable, the function to retry.
+          *args: args, array arguments to pass to the function.
+          **kwargs: kwargs, keyword arguments to pass to the function.
+        """
+        for attempt in range(self._attempts):
+            try:
+                result = func(*args, **kwargs)
+                return result
+
+            except Exception as exc:
+                # Raise the exception to caller if either is met:
+                # - We've used the last attempt.
+                # - self._retry_for is set, and we do not match.
+                # - self._do_not_retry_for is set, and we do match.
+                # - name is not actually a member of the client class.
+                if attempt >= self._attempts - 1 \
+                   or (self._retry_for
+                       and not isinstance(exc, self._retry_for)) \
+                   or (self._do_not_retry_for
+                       and isinstance(exc, self._do_not_retry_for)) \
+                   or name not in self._client_dir:
+                    raise exc
+
+                # Sleep and try again.
+                sleep(self._retry_delay)
+
+    # This is the real magic soup of the class, we catch anything that isn't
+    # strictly defined for ourselves and pass it on to whatever client we've
+    # been given.
+    def __getattr__(self, name):
+
+        return lambda *args, **kwargs: self._retry(
+            name,
+            self._client.__getattribute__(name),
+            *args,
+            **kwargs
+        )
+
+    # We implement these explicitly because they're "magic" functions and won't
+    # get passed on by __getattr__.
+
+    def __dir__(self):
+        return self._client_dir
+
+    # These magics are copied from the base client.
+    def __setitem__(self, key, value):
+        self.set(key, value, noreply=True)
+
+    def __getitem__(self, key):
+        value = self.get(key)
+        if value is None:
+            raise KeyError
+        return value
+
+    def __delitem__(self, key):
+        self.delete(key, noreply=True)
--- a/pymemcache/test/test_client_retry.py
+++ b/pymemcache/test/test_client_retry.py
@@ -1,0 +1,316 @@
+""" Test collection for the RetryingClient. """
+
+import functools
+import unittest
+
+import mock
+import pytest
+
+from .test_client import ClientTestMixin, MockSocket
+from pymemcache.client.retrying import RetryingClient
+from pymemcache.client.base import Client
+from pymemcache.exceptions import MemcacheUnknownError, MemcacheClientError
+
+
+# Test pure passthroughs with no retry action.
+class TestRetryingClientPassthrough(ClientTestMixin, unittest.TestCase):
+
+    def make_base_client(self, mock_socket_values, **kwargs):
+        base_client = Client(None, **kwargs)
+        # mock out client._connect() rather than hard-settting client.sock to
+        # ensure methods are checking whether self.sock is None before
+        # attempting to use it
+        sock = MockSocket(list(mock_socket_values))
+        base_client._connect = mock.Mock(side_effect=functools.partial(
+            setattr, base_client, "sock", sock))
+        return base_client
+
+    def make_client(self, mock_socket_values, **kwargs):
+        # Create a base client to wrap.
+        base_client = self.make_base_client(
+            mock_socket_values=mock_socket_values, **kwargs
+        )
+
+        # Wrap the client in the retrying class, disable retries.
+        client = RetryingClient(base_client, attempts=1)
+        return client
+
+
+# Retry specific tests.
+@pytest.mark.unit()
+class TestRetryingClient(object):
+
+    def make_base_client(self, mock_socket_values, **kwargs):
+        """ Creates a regular mock client to wrap in the RetryClient. """
+        base_client = Client(None, **kwargs)
+        # mock out client._connect() rather than hard-settting client.sock to
+        # ensure methods are checking whether self.sock is None before
+        # attempting to use it
+        sock = MockSocket(list(mock_socket_values))
+        base_client._connect = mock.Mock(side_effect=functools.partial(
+            setattr, base_client, "sock", sock))
+        return base_client
+
+    def make_client(self, mock_socket_values, **kwargs):
+        """
+        Creates a RetryingClient that will respond with the given values,
+        configured using kwargs.
+        """
+        # Create a base client to wrap.
+        base_client = self.make_base_client(
+            mock_socket_values=mock_socket_values
+        )
+
+        # Wrap the client in the retrying class, and pass kwargs on.
+        client = RetryingClient(base_client, **kwargs)
+        return client
+
+    # Start testing.
+    def test_constructor_default(self):
+        base_client = self.make_base_client([])
+        RetryingClient(base_client)
+
+        with pytest.raises(TypeError):
+            RetryingClient()
+
+    def test_constructor_attempts(self):
+        base_client = self.make_base_client([])
+        rc = RetryingClient(base_client, attempts=1)
+        assert rc._attempts == 1
+
+        with pytest.raises(ValueError):
+            RetryingClient(base_client, attempts=0)
+
+    def test_constructor_retry_for(self):
+        base_client = self.make_base_client([])
+
+        # Try none/default.
+        rc = RetryingClient(base_client, retry_for=None)
+        assert rc._retry_for == tuple()
+
+        # Try with tuple.
+        rc = RetryingClient(base_client, retry_for=tuple([Exception]))
+        assert rc._retry_for == tuple([Exception])
+
+        # Try with list.
+        rc = RetryingClient(base_client, retry_for=[Exception])
+        assert rc._retry_for == tuple([Exception])
+
+        # Try with multi element list.
+        rc = RetryingClient(base_client, retry_for=[Exception, IOError])
+        assert rc._retry_for == (Exception, IOError)
+
+        # With string?
+        with pytest.raises(ValueError):
+            RetryingClient(base_client, retry_for="haha!")
+
+        # With collectino of string and exceptions?
+        with pytest.raises(ValueError):
+            RetryingClient(base_client, retry_for=[Exception, str])
+
+    def test_constructor_do_no_retry_for(self):
+        base_client = self.make_base_client([])
+
+        # Try none/default.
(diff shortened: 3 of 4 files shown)


pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # This is the real magic soup of the class, we catch anything that isn't strictly defined for
+    # ourselves and pass it on to whatever client we've been given.
+    def __getattr__(self, name, *args, **kwargs):
+        # For no retry:
+        # return self.client.__getattribute__(name, *args, **kwargs)
+        for attempt in range(self.attempts):
+            try:
+                return self.client.__getattribute__(name, *args, **kwargs)
+            except Exception as exc:
```
cgordon, line 69: Have a special case for `AttributeError` in case the user calls a function that isn't defined on the underlying `client`.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+    # Check that all the elements are actually inherited from Exception. (Catchable)
+    if not all([issubclass(arg, Exception) for arg in argument_tuple]):
+        raise ValueError("%s contained elements other than Exceptions." % argument_name)
+
+class RetryingClient(object):
+    """
+    Client that allows retrying calls for the other clients.
+    """
+
+    def __init__(
```
cgordon, line 46: Document the arguments to this method. For example, what does it mean for `retry_for` to be `None`? Does it mean that all exceptions cause retries, or no exceptions cause retries?

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+        self.retry_delay = retry_delay
+        self.retry_for = _ensure_tuple_argument("retry_for", retry_for)
+        self.do_not_retry_for = _ensure_tuple_argument("do_not_retry_for", do_not_retry_for)
+
+
+    # This is the real magic soup of the class, we catch anything that isn't strictly defined for
+    # ourselves and pass it on to whatever client we've been given.
+    def __getattr__(self, name, *args, **kwargs):
+        # For no retry:
+        # return self.client.__getattribute__(name, *args, **kwargs)
```
cgordon, line 65: I would just leave this comment out, since it doesn't add anything to the understanding.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+        for attempt in range(self.attempts):
+            try:
+                return self.client.__getattribute__(name, *args, **kwargs)
+            except Exception as exc:
+
+                if attempt == self.attempts:
+                    # If we're at the limit, just raise the exception to caller.
+                    # Exception type doesn't matter.
+                    raise exc
+                elif isinstance(exc, self.retry_for) or not isinstance(exc, self.do_not_retry_for):
```
cgordon, line 75: There are four cases here:

(1) exc in self.retry_for, exc NOT in self_do_not_retry_for: this case is correct, and will retry

(2) exc in self.retry_for, exc in self.do_not_retry_for: this will retry the exception, and this is probably a user error, maybe check that users don't do this as a precondition in __init__?

(3) exc NOT in self.retry_for, exc in self.do_not_retry_for: this case is correct, and will not retry

(4) exc NOT in self.retry_for, exc NOT in self.do_not_retry_for: this case is sort of questionable. It will retry the exception, even though the user didn't explicitly ask for it to be retried. This is where we need a more clear contract for what users can expect from this logic.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+    # ourselves and pass it on to whatever client we've been given.
+    def __getattr__(self, name, *args, **kwargs):
+        # For no retry:
+        # return self.client.__getattribute__(name, *args, **kwargs)
+        for attempt in range(self.attempts):
+            try:
+                return self.client.__getattribute__(name, *args, **kwargs)
+            except Exception as exc:
+
+                if attempt == self.attempts:
```
cgordon, line 71: I think you can simplify this conditional block as:

```
if attempt == self.attempts or not isinstance(exc, self._retry_for) or isinstance(exc, self.do_not_retry_for):
    raise exc

sleep(self.retry_delay)
```

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,297 @@
...
+        value = self.get(key)
+        if value is None:
+            raise KeyError
+        return value
+
+    def __delitem__(self, key):
+        return self.delete(key, noreply=True)
+
+
+    # "static" implementations.
```
cgordon, line 104: You can just delete the comments below this line and stick with the dynamic implementation.

cgordon (changes requested):
There are a couple existing "retry" libraries on PyPI, and it is probably worth either implementing their full set of functionality, or just using them directly. Here are the two I found:

https://pypi.org/project/retry/
https://pypi.org/project/retrying/

pymemcache/client/retrying.py
martinnj: Just to be sure, you mean catch `AttributeError`?
Propegating it seems like the right call since you've essentially tried to call a specific method from the underlying `client`, that method doesn't exist. Which would also give the `AttributeError` if called directly.

Or did I misunderstand the comment? :)

pymemcache/client/retrying.py
martinnj: This one will be clarified with documentation. Case 4 is a little dicey yes.
My intuition was that if you haven't  asked specified any go/no-go exceptions, revert to simpler behavior: Everything is just getting retried.

pymemcache/client/retrying.py
martinnj: Nice catch. Thanks. 👍

pymemcache/client/retrying.py
cgordon: We need to make sure to never retry on an `AttributeError`.

cgordon: Of course, a user of the library should be able to retry on their own `AttributeError`s! So we need logic that ignores AttributeError thrown by our code only.

martinnj:
> 
> 
> There are a couple existing "retry" libraries on PyPI, and it is probably worth either implementing their full set of functionality, or just using them directly. Here are the two I found:
> 
> https://pypi.org/project/retry/
> https://pypi.org/project/retrying/

Was there any specific features from those libraries you where wanting?
Currently the timeout, retry count and exception based actions is implemented.

I didn't look for existing libraries because adding dependencies for something like this would be a shame in my eyes.

pymemcache/client/retrying.py
martinnj: Gotcha!

4383:
> There are a couple existing "retry" libraries on PyPI, and it is probably worth either implementing their full set of functionality, or just using them directly. Here are the two I found:
> 
> https://pypi.org/project/retry/
> https://pypi.org/project/retrying/

Tenacity is an interesting implementation and already well used on openstack.

However, I agree with @martinnj , if possible, it would be more interesting to keep pymemcache as a pure python library (I voluntarly ignore `six`). Maybe by only providing specific retry strategies without tons of configurations. Less is more.

4383:
Also notice that Wednesday I also started to implement similar features, however, I didn't yet proposed them before today  through a pull request.  #326

By introspection the passed clients, I dynamically decorate their methods to make them "retryable".

It's not a problem for me if you decide to continue with this version (that of @martinnj). Maybe we could converge our implementations to design and co-author something, it's up to you to decide. I propose you dispose.

martinnj:
> It's not a problem for me if you decide to continue with this version (that of @martinnj). Maybe we could converge our implementations to design and co-author something, it's up to you to decide. I propose you dispose.

Wild timing, I'll just wait with spending more time until a collaborator decides either way. :)

pymemcache/client/retrying.py
martinnj: Redid the retry logic so this isn't relevant anymore. :)

martinnj:
I had some time so I thought I'd give it a shot anyway.

Changes:
- I had to redo the retry/passthrough logic since I had grossly misunderstood `__getattr__`.
- Fixed style issues.
- Fleshed out docstrings.
- Added unit tests.

I haven't added a unit-test for the "delay" function. If forced I would time how long the call took and check that it was strictly longer than the delay argument. But I also feel like inserting second long delays in the unit-tests is a bit weird.

cgordon:
@martinnj I'm out of town right now and mostly away from my computer. I will have time to look through this next week, apologies for the delay.

martinnj:
> 
> 
> @martinnj I'm out of town right now and mostly away from my computer. I will have time to look through this next week, apologies for the delay.

No stress, it's all good.

4383:
Hello

@martinnj: Please can you also add related doc with some examples of usage? 
maybe there https://pymemcache.readthedocs.io/en/latest/getting_started.html

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,183 @@
...
+      ValueError: If the argument was not None, tuple or Iterable.
+      ValueError: If any of the elements of the argument is not a subclass of
+        Exception.
+    """
+
+    argument_tuple = None
+
+    # Ensure the argument is a tuple, or raise an error if we can't convert it.
+    if argument is None:
+        return tuple()
```
4383, line 31: This could be the default value of `argument_tuple` and the condition block could be reduced to if/elif/else:

Example:
```
if isinstance(argument, tuple):
        argument_tuple = argument
elif isinstance(argument, Iterable) and \
            not isinstance(argument, (str, bytes, bytearray)):
        argument_tuple = tuple(argument)
else:
        raise ValueError("%s was not a tuple or Iterable." % argument_name)
```

Anyway, I don't want to nitpick our current implementation, this code LGTM in the current state.

4383 (changes requested):
Hello see my inline comment about abc

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Workhorse function, handles retry logic.
+
+        Args:
+          name: str, Name of the function called.
+          func: callable, the function to retry.
+          *args: args, array arguments to pass to the function.
+          **kwargs: kwargs, keyword arguments to pass to the function.
+        """
+        for attempt in range(1, self._attempts + 1):
```
4383, line 132: Can't we start our range from `0` to `self._attempts` to avoid to bump `self._attempts`?

By example if user pass `attempts=1` that will avoid a mental gymnastic to get our starting point during code reading. You protected the execution from values inferior to `1`, so I don't think that we need to tweak it during runtime.

pymemcache/test/test_client_retry.py
```diff
@@ -0,0 +1,315 @@
+
+import functools
```
4383, line 2: From a pep8 point of view I think that you could group the imports by lib types (stdlib first, then third libraries, and then the current lib)

pymemcache/client/retrying.py
4383: This check can remain. It will be complementary with testing list and tuple. See my next comments.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,183 @@
+""" Module containing the RetryingClient wrapper class. """
+
+from collections import Iterable
```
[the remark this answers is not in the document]
martinnj: The library currently supports Python 2.7 which doesn't have `collections.abc`, but I don't see any reason why it can't be done with a `try .. except`. 👍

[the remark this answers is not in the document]
4383: Oh yes good catch. My bad.
I was thinking that the py2 support was already dropped, but indeed that's not the case yet

https://github.com/pinterest/pymemcache/pull/321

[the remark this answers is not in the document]
martinnj: Import became unnecessary after the check simplification below, so just removed it. Nicely spotted.

pymemcache/client/retrying.py
martinnj: Went ahead and simplified the check as discussed below.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    argument_tuple = None
+
+    # Ensure the argument is a tuple, or raise an error if we can't convert it.
+    if argument is None:
+        return tuple()
+    elif isinstance(argument, tuple):
+        argument_tuple = argument
+    elif isinstance(argument, Iterable) and \
```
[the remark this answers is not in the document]
martinnj: Yeah, simplified the check. 
https://github.com/martinnj/pymemcache/commit/5da40908d0c9628e6185ca119bcd98afa65e279c

pymemcache/test/test_client_retry.py
martinnj: You're right. Didn't clean up properly after myself when I was writing tests.
https://github.com/martinnj/pymemcache/commit/03d7c48fba36934659fc37f1a4844da46d00f681

pymemcache/client/retrying.py
martinnj: Sure. I'm am not opinionated on it.
https://github.com/martinnj/pymemcache/commit/b0782c595dafa92c1608abd8f749c255ebf71edd

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,178 @@
...
+      ValueError: If the argument was not None, tuple or Iterable.
+      ValueError: If any of the elements of the argument is not a subclass of
+        Exception.
+    """
+
+    # Ensure the argument is a tuple, set or list.
+    if argument is None:
+        return tuple()
+    elif not isinstance(argument, (tuple, set, list)):
+        raise ValueError("%s was not a tuple or Iterable." % argument_name)
```
4383, line 30: Sorry I didn't noticed this exception message during my previous review.

What do you think about a message like this:

> An error occur with your passed argument. This method accept three possible kind of sequence type (tuple, set, list) for the argument "argument". If you see this message that mean that your argument isn't valide. Your argument type is {type(argument)}. If you continue to see this message, then, please check your passed arguments.

Maybe we could rename "argument" as "exceptions" and rename "argument_name" as "exception_name".
In this context I think that we could speak about exception rather than seeking to be too generalist. That will make the code more human readable, explicit and semantic.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # Convert the argument before checking contents.
+    argument_tuple = tuple(argument)
+
+    # Check that all the elements are actually inherited from Exception.
+    # (Catchable)
+    if not all([issubclass(arg, Exception) for arg in argument_tuple]):
+        raise ValueError(
+            "%s contained elements other than Exceptions." % argument_name
```
4383, line 39: Some kind of remark here.

The error message could be more detailed and it could bring more details about why the users face it and what are the solutions for them. We should be proactive and lead users on the right path to avoid mistakes and so reduce the human support in a second time (github issues, etc...).

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,178 @@
...
+        self._retry_for = _ensure_tuple_argument("retry_for", retry_for)
+        self._do_not_retry_for = _ensure_tuple_argument(
+            "do_not_retry_for", do_not_retry_for
+        )
+
+        # Verify no overlap in the go/no-go exception collections.
+        for exc_class in self._retry_for:
+            if exc_class in self._do_not_retry_for:
+                raise ValueError(
+                    "%s was present in both `retry_for` and "
```
4383, line 110: Same think here. We should define in the message why the users face it, what we except and what users should do in this case.

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,178 @@
...
+          ValueError: If `retry_for` or `do_not_retry_for` is not None, tuple or
+            Iterable.
+          ValueError: If any of the elements of `retry_for` or
+            `do_not_retry_for` is not a subclass of Exception.
+          ValueError: If there is any overlap between `retry_for` and
+            `do_not_retry_for`.
+        """
+
+        if attempts < 1:
+            raise ValueError("`attempts` argument must be at least 1.")
```
4383, line 96: here too

4383 (commented):
Nice changes, thank you for applying my previous comments.

Here is a couple of new suggestions, they aren't mandatory comments. I think that they can help to improve this feature.

4383:
Also, I think that it could be worth to reword and to squash your commits into a all in one commit to avoid to spread the history of these changes.

Example:
```
$ git rebase -i HEAD~12
reword ee1aa4c Add first draft for RetryingClient.
squash 9325a3c Corrections, docstrings and fixes for RetryingClient.
squash 0700983 Stop retring AttributeError and style fixes.
squash eddbe50 Stylefixes for flake8.
squash 6664775 Minor fixes and docstrings for RetryingClient.
squash cf2b44d Add unit-tests for RetryingClient.
squash f08d348 Add RetryingClient example to the getting started docuementation.
squash 71f81bb More robust import of Iterable.
squash b0782c5 0 index retry range.
squash 03d7c48 Reorder imports of tests for pop8ness.
squash eee78cc Sharpen test-case.
squash 5da4090 Simplify argument parsing.
```

All these changes are related to a single one feature.

docs/getting_started.rst
```diff
@@ -77,6 +77,34 @@ on if a server goes down.
     client.set('some_key', 'some value')
     result = client.get('some_key')
 
+Using a built-in retrying
+-------------------------
+The library comes with a retrying client that allows you to retry calls that
```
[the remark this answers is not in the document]
martinnj: I'm not strongly opinionated on documentation. In my experience if someone who didn't write the code tells you that a documentation part is easier to understand in a different way, they're usually right. :)
Done in: https://github.com/martinnj/pymemcache/commit/8c6013e04c7410430d06186cc7b0e92b4486c14d

pymemcache/client/retrying.py
martinnj: I don't know that I agree with having an exception message that long. Makes it seem like the information should be in the client docstring (both the check method and the constructor) instead.

I agree that the wording could perhaps be better, and I also see that the exception message doesn't actually match the types we check for.

My reasoning for the variable naming is like so:
- `argument_name` is the name of the argument we're currently checking. In itself it doesn't have anything to do with an exception, it's just for giving sensible `ValueError` messages, it will tell whoever catches it exactly which argument (`retry_for` or `do_not_retry_for`) contained an invalid value.
- `argument` is the value we're checking. I'm renaming it to `argument_value` since this is more explicit.

Done in https://github.com/martinnj/pymemcache/commit/6bd21d0261f57d54c8c3d1211088bc27f07f5ee3

pymemcache/client/retrying.py
martinnj: I'm still not sold on super wordy exception messages, so I'll let a collaborator weigh in on that one.
I've reworded the exception to be more specific about what is required though.
https://github.com/martinnj/pymemcache/commit/6bd21d0261f57d54c8c3d1211088bc27f07f5ee3

pymemcache/client/retrying.py
martinnj: Somewhat done with https://github.com/martinnj/pymemcache/commit/d01efbb09cdb5bc4d3d62eb59df714256833c2bd

pymemcache/client/retrying.py
martinnj: Somewhat done with https://github.com/martinnj/pymemcache/commit/d01efbb09cdb5bc4d3d62eb59df714256833c2bd

martinnj:
> 
> 
> Also, I think that it could be worth to reword and to squash your commits into a all in one commit to avoid to spread the history of these changes.
> 
> Example:
> 
> ```
> $ git rebase -i HEAD~12
> reword ee1aa4c Add first draft for RetryingClient.
> squash 9325a3c Corrections, docstrings and fixes for RetryingClient.
> squash 0700983 Stop retring AttributeError and style fixes.
> squash eddbe50 Stylefixes for flake8.
> squash 6664775 Minor fixes and docstrings for RetryingClient.
> squash cf2b44d Add unit-tests for RetryingClient.
> squash f08d348 Add RetryingClient example to the getting started docuementation.
> squash 71f81bb More robust import of Iterable.
> squash b0782c5 0 index retry range.
> squash 03d7c48 Reorder imports of tests for pop8ness.
> squash eee78cc Sharpen test-case.
> squash 5da4090 Simplify argument parsing.
> ```
> 
> All these changes are related to a single one feature.

Plan was to squash just before merging, since the commits give context to some of the comments in the PR.

pymemcache/client/retrying.py
martinnj: Also Somewhat done with https://github.com/martinnj/pymemcache/commit/d01efbb09cdb5bc4d3d62eb59df714256833c2bd

pymemcache/client/retrying.py
4383: No problem, the current version LGTM, that was more a suggestion. Leave that as it is if you think that it's ok.

4383:
> > Also, I think that it could be worth to reword and to squash your commits into a all in one commit to avoid to spread the history of these changes.
> > Example:
> > ```
> > $ git rebase -i HEAD~12
> > reword ee1aa4c Add first draft for RetryingClient.
> > squash 9325a3c Corrections, docstrings and fixes for RetryingClient.
> > squash 0700983 Stop retring AttributeError and style fixes.
> > squash eddbe50 Stylefixes for flake8.
> > squash 6664775 Minor fixes and docstrings for RetryingClient.
> > squash cf2b44d Add unit-tests for RetryingClient.
> > squash f08d348 Add RetryingClient example to the getting started docuementation.
> > squash 71f81bb More robust import of Iterable.
> > squash b0782c5 0 index retry range.
> > squash 03d7c48 Reorder imports of tests for pop8ness.
> > squash eee78cc Sharpen test-case.
> > squash 5da4090 Simplify argument parsing.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > All these changes are related to a single one feature.
> 
> Plan was to squash just before merging, since the commits give context to some of the comments in the PR.
> 
> Edit:
> Thanks for all the feedback btw. :)

Thanks to you too :)

No problem it make sense to squash it just before.

martinnj:
@cgordon Hate to be pushy, but any news on this one? :)

4383 (approved):

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,185 @@
...
+            Accepts any class that is a subclass of Exception.
+            Defaults to None.
+
+          do_not_retry_for: optional None|tuple|set|list, what
+            exceptions should be retried. Will not block retries for any
+            Exception if None.
+            Example:
+                `(IOError, MemcacheIllegalInputError)`
+            Accepts any class that is a subclass of Exception.
+            Defaults to None.
```
cgordon, line 86: I think it would be nicer to take a function that takes an `Exception` as an argument and returns `True` to retry or `False` to stop retries. That will give clients a lot more flexibility, and make the logic a little more straightforward. This is a subjective thing, so I won't hold up the commit on it.

cgordon (commented):
LGTM! I left one comment that you can do, or not, at your discretion.

cgordon:
Hmm, I hit "Comment" instead of "Approve" by accident, anyone know how to navigate this UI to approve this PR?

pymemcache/client/retrying.py
martinnj: I agree that passing something callable in would make it more flexible. But ff you are not strongly opinionated on this I would like to keep it like this, since it makes invocation simpler.

It might be an feature to add later though. I don't see them having to exclude each other, as long as any priorities in the checks is in the documentation.

martinnj:
> 
> 
> Hmm, I hit "Comment" instead of "Approve" by accident, anyone know how to navigate this UI to approve this PR?

Good question. No clue. I've hit "Re-request review" for the blocking changes, might give you a prompt to clear it. :)

cgordon (approved):

PR #326, opened 2021-06-18
[WIP] Adding retry management

4383:
Still work in progress.

The goal is to decorate all passed client dynamically to avoid technical
debt and API changes.

This module will only decorate passed client to manage uniformized and
standardized retries. It will be possible to configure the retry
strategies and mechanismes.

Here is just a PoC for now.

### DIFF
--- a/pymemcache/client/retrying.py
+++ b/pymemcache/client/retrying.py
@@ -1,0 +1,183 @@
+import functools
+import inspect
+import time
+import logging
+
+import pymemcache.exceptions as pymemexception
+
+logger = logging.getLogger(__name__)
+
+default_strategies = {
+    'global'
+    'local'
+}
+
+
+class _retry_always(retry_base):
+    """Retry strategy that always rejects any result."""
+
+    def __call__(self, retry_state):
+        return True
+
+
+retry_always = _retry_always()
+
+
+class retry_if_exception(retry_base):
+    """Retry strategy that retries if an exception verifies a predicate."""
+
+    def __init__(self, predicate):
+        self.predicate = predicate
+
+    def __call__(self, retry_state):
+        if retry_state.outcome.failed:
+            return self.predicate(retry_state.outcome.exception())
+        else:
+            return False
+
+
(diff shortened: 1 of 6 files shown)


4383:
Related to #307

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+import pymemcache.exceptions as pymemexception
+
+logger = logging.getLogger(__name__)
+
+def retry(func):
+    @functools.wraps(func)
+    def func_call(*args, **kwargs):
+        ntries=15
```
4383, line 15: Still WIP so hard coded, but the goal is to provide some configurable options (backoff, number of retry, duration, etc...)

pymemcache/client/retrying.py
```diff
@@ -0,0 +1,48 @@
...
+        while ntries:
+            try:
+                return func(*args, **kwargs)
+            except pymemexception.MemcacheUnexpectedCloseError as e:
```
4383, line 19: Same thing here, I think that we need to provide strategies. Either we allow handle all the kinds of exceptions, or only the pymemcache exceptions, or again only one specific type of exception passed by user.

For now it's just a dumb PoC for my own usage.

pymemcache/client/retrying.py
4383: We need more robust strategy, like a backoff strategy with a range of possible value to avoid to flood the server and manage system workload more finely. If many clients try to reconnect at the same time the backoff strategy will help to spread the load.

4383 (commented):
I propose to converge with #324 to co-author something to:
- avoid duplicated reviews and stuffs;
- design something standardized.

4383:
**Fresh paint! Does not yet work, please don't spend time on it for now.**

4383:
I abandoning these changes to give us the focus on #324 . @martinnj made significant changes that seems to works. Let's give him the priority.

Thanks for our attention.

PR #328, opened 2021-06-18
Client API for server shutdown protocol command

LINKIWI:
This change proposes the addition of a client API wrapper for the [`shutdown` text protocol command](https://github.com/memcached/memcached/blob/c472369fed5981ba8c004d426cee62d5165c47ca/doc/protocol.txt#L1698).

Minimal example usage:

```python


c = Client('localhost:11211')
c.shutdown(False)
```

When shutdown is not enabled on the server:

```
<28 new auto-negotiating client connection
28: Client using the ascii protocol
<28 shutdown
>28 ERROR: shutdown not enabled
<28 connection closed.
```

```
  File "driver.py", line 5, in <module>
    c.shutdown(False)
  File "/home/kiwi/sync/code/external/pymemcache/pymemcache/client/base.py", line 845, in shutdown
    self._misc_cmd([cmd], b'shutdown', False)
  File "/home/kiwi/sync/code/external/pymemcache/pymemcache/client/base.py", line 1045, in _misc_cmd
    self._raise_errors(line, cmd_name)
  File "/home/kiwi/sync/code/external/pymemcache/pymemcache/client/base.py", line 853, in _raise_errors
    raise MemcacheUnknownCommandError(name)
pymemcache.exceptions.MemcacheUnknownCommandError: b'shutdown'
```

When [`--enable-shutdown`](https://github.com/memcached/memcached/blob/c472369fed5981ba8c004d426cee62d5165c47ca/memcached.c#L4749) is specified at server runtime:

```
<28 new auto-negotiating client connection
28: Client using the ascii protocol
<28 shutdown
Signal handled: Interrupt.
<28 connection closed.
stopped assoc
asking workers to stop
asking background threads to stop
stopped lru crawler
stopped maintainer
stopped slab mover
stopped logger thread
stopped idle timeout thread
closing connections
<26 connection closed.
<27 connection closed.
reaping worker threads
all background threads stopped
```

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -817,6 +817,35 @@ class Client(object):
         self._misc_cmd([cmd], b'quit', True)
         self.close()
 
+    def shutdown(self, graceful=False):
+        """
+        The memcached "shutdown" command.
+
+        This will request shutdown and eventual termination of the server,
+        optionally preceded by a graceful stop of memcached's internal state
+        machine. Note that the server needs to have been started with the
+        shutdown protocol command enabled with the --enable-shutdown flag.
+
+        Args:
+          graceful: optional bool, True to request a graceful shutdown with
+                    SIGUSR1 (defaults to False, i.e. SIGINT shutdown).
+
+        Returns:
+          True.
+        """
+        cmd = b'shutdown'
+        if graceful:
+            cmd += b' graceful'
+        cmd += b'\r\n'
+
+        # The shutdown command raises a server-side error if the shutdown
+        # protocol command is not enabled. Otherwise, a successful shutdown
+        # is expected to close the remote end of the transport.
+        try:
+            self._misc_cmd([cmd], b'shutdown', False)
+        except MemcacheUnexpectedCloseError:
+            return True
+
     def _raise_errors(self, line, name):
         if line.startswith(b'ERROR'):
             raise MemcacheUnknownCommandError(name)
@@ -1248,6 +1277,10 @@ class PooledClient(object):
                 client.quit()
             finally:
                 self.client_pool.destroy(client)
+
+    def shutdown(self, graceful=False):
+        with self.client_pool.get_and_release(destroy_on_fail=True) as client:
+            client.shutdown(graceful)
 
     def __setitem__(self, key, value):
         self.set(key, value, noreply=True)
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -829,9 +829,6 @@ class Client(object):
         Args:
           graceful: optional bool, True to request a graceful shutdown with
                     SIGUSR1 (defaults to False, i.e. SIGINT shutdown).
-
-        Returns:
-          True.
         """
         cmd = b'shutdown'
         if graceful:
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -817,6 +817,37 @@ def quit(self):
...
+
+        # The shutdown command raises a server-side error if the shutdown
+        # protocol command is not enabled. Otherwise, a successful shutdown
+        # is expected to close the remote end of the transport.
+        try:
+            self._misc_cmd([cmd], b'shutdown', False)
+        except MemcacheUnexpectedCloseError:
+            pass
+
+        return True
```
jparise, line 849: Perhaps this function should not return anything (`None` by default), which would match `quit`'s behavior.

pymemcache/client/base.py
```diff
@@ -1249,6 +1280,10 @@ def quit(self):
             finally:
                 self.client_pool.destroy(client)
 
+    def shutdown(self, graceful=False):
+        with self.client_pool.get_and_release(destroy_on_fail=True) as client:
+            client.shutdown(graceful)
```
jparise, line 1285: Should the client be destroyed on a successful shutdown?

```suggestion
            client.shutdown(graceful)
            self.client_pool.destroy(client)
```

jparise (commented):
Thanks for contributing this! Just a few questions about the implementation because I haven't used memcached's `shutdown` support.

pymemcache/client/base.py
jparise: I suppose it could also return `False` if `MemcacheUnknownCommandError` is raised, but I think it's better to surface that error directly to the caller.

pymemcache/client/base.py
LINKIWI: I'm not sure; I guess it depends on the desired interface contract. The changeset as it is now proposes that this API be a relatively thin wrapper for server shutdown, i.e. changing the state of the server without changing the state in the client (tearing down the connection). I feel that destroying the client (modifying the client state) should be done explicitly at the callsite with an subsequent `quit` or similar.

The semantics of this new API could also be changed to mean "invoke both client and server shutdown" but that conflates the responsibilities I think. For example, imagine the case where you deploy memcached behind a protocol-aware proxy (which we do); you might not necessarily want the client<>proxy connection destroyed after relaying the shutdown command between proxy<>server(s).

jparise: Yes, I generally agree with you.

Because I'm not familiar with `shutdown`, could you say what happens from the client's perspective after a successful shutdown is issued? Is the client still usable (perhaps after a reconnect)?

LINKIWI: In general, the client should be usable after a reconnect. For the direct pymemcache <> server connection use case, I would say there likely isn't any practical reason to keep the local connection valid, but it's more a consideration for isolating the behavior of the interface (opting not to modify local state in a method designed to modify server state).

Another option could be adding a kwarg like `quit` or `close` whose default is `True` which optionally controls destroying the connection after successful shutdown. This could support use cases where one would want to manage the connection lifecycle manually/deploy behind a proxy/etc. while also supporting the more common use case.

jparise: I think it's fine for clients who also want the `quit` behavior to call `quit()` explicitly after `shutdown()`.

jparise (approved):

PR #332, opened 2021-06-29
Implement socket keepalive configuration mechanisms

4383:
These changes allow users to activate and configure their socket keepalive.

For now users can configure:
- the enabling of their socket keepalive;
- the time the connection needs to remain idle before TCP starts sending
  keepalive probes;
- the time (in seconds) between individual keepalive probes;
- the maximum number of keepalive probes TCP should send before dropping
  the connection.

This could be used by users who want to handle fine grained failures and
scenarios.

Also this could be another way to handle problems more or less similar
to https://github.com/pinterest/pymemcache/issues/307

This could be in addition of the retry mechanisms
https://github.com/pinterest/pymemcache/pull/324

For now this feature only support Linux platform.
Other systems support could be added through follow up patches.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -348,7 +348,7 @@ class Client(object):
             if not isinstance(self.socket_keepalive, KeepaliveOpts):
                 raise ValueError(
                     "Unsupported keepalive options. If you see this message "
-                    "it mean that you passed an unsupported object within "
+                    "it means that you passed an unsupported object within "
                     "the param `socket_keepalive`. To fix it "
                     "please instantiate and pass to socket_keepalive a "
                     "KeepaliveOpts object. That's the only supported type "
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import errno
+import platform
 import socket
 import six
 
@@ -36,6 +37,10 @@ VALID_STORE_RESULTS = {
     b'append':  (b'STORED', b'NOT_STORED'),
     b'prepend': (b'STORED', b'NOT_STORED'),
     b'cas':     (b'STORED', b'EXISTS', b'NOT_FOUND'),
+}
+
+SOCKET_KEEPALIVE_SUPPORTED_SYSTEM = {
+    'Linux',
 }
 
 
@@ -128,6 +133,44 @@ def normalize_server_spec(server):
     if host.startswith('['):
         host = host.strip('[]')
     return (host, port)
+
+
+class KeepaliveOpts(object):
+    """
+    A configuration structure to define the socket keepalive.
+
+    This structure must be passed to a client. The client will configure
+    its socket keepalive by using the elements of the structure.
+    """
+    __slots__ = ('idle', 'intvl', 'cnt')
+
+    def __init__(self, idle=1, intvl=1, cnt=5):
+        """
+        Constructor.
+
+        Args:
+          idle: The time (in seconds) the connection needs to remain idle
+            before TCP starts sending keepalive probes. Should be a positive
+            integer most greater than zero.
+          intvl: The time (in seconds) between individual keepalive probes.
+            Should be a positive integer most greater than zero.
+          cnt: The maximum number of keepalive probes TCP should send before
+            dropping the connection. Should be a positive integer most greater
+            than zero.
+        """
+
+        if idle < 1:
+            raise ValueError(
+                "The idle parameter must be greater or equal to 1.")
+        self.idle = idle
+        if intvl < 1:
+            raise ValueError(
+                "The intvl parameter must be greater or equal to 1.")
+        self.intvl = intvl
+        if cnt < 1:
+            raise ValueError(
+                "The cnt parameter must be greater or equal to 1.")
+        self.cnt = cnt
 
 
 class Client(object):
@@ -236,6 +279,7 @@ class Client(object):
                  no_delay=False,
                  ignore_exc=False,
                  socket_module=socket,
+                 socket_keepalive=None,
                  key_prefix=b'',
                  default_noreply=True,
                  allow_unicode_keys=False,
@@ -262,6 +306,9 @@ class Client(object):
             misses. Defaults to False.
           socket_module: socket module to use, e.g. gevent.socket. Defaults to
             the standard library's socket module.
+          socket_keepalive: Activate the socket keepalive feature by passing
+            a KeepaliveOpts structure in this parameter. Disabled by default
+            (None). This feature is only supported on Linux platforms.
           key_prefix: Prefix of key. You can use this as namespace. Defaults
             to b''.
           default_noreply: bool, the default value for 'noreply' as passed to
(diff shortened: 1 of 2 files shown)


pymemcache/client/base.py
```diff
@@ -38,6 +39,10 @@
     b'cas':     (b'STORED', b'EXISTS', b'NOT_FOUND'),
 }
 
+SOCKET_KEEPALIVE_SUPPORTED_SYSTEM = [
+    'Linux',
+]
```
jparise, line 44: Let's use a `set` here:

```suggestion
SOCKET_KEEPALIVE_SUPPORTED_SYSTEM = {
    'Linux',
}
```

pymemcache/client/base.py
```diff
@@ -281,6 +298,25 @@ def __init__(self,
...
+        if self.socket_keepalive:
+            if user_system not in SOCKET_KEEPALIVE_SUPPORTED_SYSTEM:
+                raise SystemError(
+                    "Pymemcache its socket keepalive mechanisms doesn't "
+                    "support your system ({user_system}). If "
+                    "you see this message that mean that you tried to "
+                    "configure your socket keepalive on an unsupported "
+                    "system. To fix the problem pass `socket_"
+                    "keepalive=False` or use a supported system. "
+                    "Supported systems are: {systems}".format(
```
jparise, line 315:
```suggestion
                    "Pymemcache's socket keepalive mechaniss doesn't "
                    "support your system ({user_system}). If "
                    "you see this message, it means that you tried to "
                    "configure your socket keepalive on an unsupported "
                    "system. To fix the problem, pass `socket_"
                    "keepalive=False` or use a supported system. "
                    "Supported systems are: {systems}".format(
```

pymemcache/client/base.py
```diff
@@ -281,6 +298,25 @@ def __init__(self,
...
+                        user_system=user_system,
+                        systems=", ".join(SOCKET_KEEPALIVE_SUPPORTED_SYSTEM)
```
jparise, line 317:
```suggestion
                        systems=", ".join(sorted(SOCKET_KEEPALIVE_SUPPORTED_SYSTEM))
```

pymemcache/client/base.py
```diff
@@ -236,6 +241,10 @@ def __init__(self,
                  no_delay=False,
                  ignore_exc=False,
                  socket_module=socket,
+                 socket_keepalive=False,
+                 socket_keepaliveidle=1,
+                 socket_keepaliveintvl=1,
+                 socket_keepalivecnt=5,
```
jparise, line 247: Would these read better with an extra underscore?

```suggestion
                 socket_keepalive_idle=1,
                 socket_keepalive_intvl=1,
                 socket_keepalive_cnt=5,
```

Alternatively, because these are only relevant for keepalive, maybe we should define a single configuration structure and use that here instead.

```py
class KeepaliveOpts(object):
    __slots__ = ('idle', 'intvl', 'cnt')

    def __init__(self, idle=1, intvl=1, cnt=5):
        self.idle = idle
        self.intvl = intvl
        self.cnt = cnt
```

Or simply a dictionary of options:

```py
# Defaults:
SOCKET_KEEPALIVE_DEFAULTS = {
    socket.TCP_KEEPIDLE: 1,
    socket.TCP_KEEPINTVL: 1,
    socket.TCL_KEEPCNT: 5,
}

# Override based on provided argument (socket_keepalive_opts):

self.socket_keepalive_opts = SOCKET_KEEPALIVE_DEFAULTS
self.socket_keepalive_opts.update(socket_keepalive_opts)
```

What do you think?

4383:
@jparise: The latest patch set address your previous comments. I refactored my code to use a structure as you suggested, I think that's a good idea.

pymemcache/client/base.py
```diff
@@ -130,6 +135,56 @@ def normalize_server_spec(server):
     return (host, port)
 
 
+def get_positive_integer(param):
+    """
+    Check if the passed param is a positive integer and return an integer
+    from it.
+
+    Zero, negative value, or non numerical values are not allowed.
+    String representation of a number can be passed.
```
jparise, line 144: As we look toward type hints in a future version, I think we want to start being stricter about which types we accept, especially for new code. I therefore think we should only accept integers here (and not strings or floats).

Given that, we can move the positive integer checks into `KeepaliveOpts.__init__` below and remove this helper function.

pymemcache/client/base.py
```diff
@@ -262,6 +318,9 @@ def __init__(self,
             misses. Defaults to False.
           socket_module: socket module to use, e.g. gevent.socket. Defaults to
             the standard library's socket module.
+          socket_keepalive: Activate the socket keepalive feature by passing
+            a KeepaliveOpts structure in parameter. Disabled by default
```
jparise, line 322:
```suggestion
            a KeepaliveOpts structure in this parameter. Disabled by default
```

pymemcache/client/base.py
```diff
@@ -262,6 +318,9 @@ def __init__(self,
...
+            (None). This feature only support Linux platforms.
```
jparise, line 323:
```suggestion
            (None). This feature is only supported on Linux platforms.
```

jparise (commented):
Looks good! Just a few remaining comments.

pymemcache/client/base.py
4383: Make sense, however, what do you think about simply moving this function as a method of the `KeepaliveOpts` class rather than moving the check into the constructor?

It will avoid to duplicate the check, it will reduce the technical debt and it will reduce the number of line.

This is a compromising between the previous version and the future version.

jparise: But it still allows non-`int` values? What is the reason why we need to provide that type conversion for new code?

4383: The goal isn't to allow non-int values, the goal is the contrary. Force the type.
We only ensure that we pass the right type of value else it will fail later during the moment when we set the socket option, example:

```
>>> import socket
>>> sock = socket.socket(socket.SOCK_STREAM)
>>> sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, "1")
TypeError: a bytes-like object is required, not 'str'
```

Ensuring the passed type is right here is cheap. It could be easily removed later when you will enforce strict type.

Else, we could remove my type checking as you suggested and we could rely only on the `TypeError` raised in the example above. It will inform the user that something went wrong. It is more a pythonic way to handle that. I'm not against too.

Thoughts?

jparise: Yes, I'd rather keep our end of the code minimal and specific for these cases (without the type conversions).

4383: Ok no problem, then I'll submit a new patch set soon.

4383: Done with fc4df8c

pymemcache/client/base.py
```diff
@@ -281,6 +328,32 @@ def __init__(self,
...
+                    "Supported systems are: {systems}".format(
+                        user_system=user_system,
+                        systems=", ".join(sorted(
+                            SOCKET_KEEPALIVE_SUPPORTED_SYSTEM))
+                    )
+                )
+            if not isinstance(self.socket_keepalive, KeepaliveOpts):
+                raise ValueError(
+                    "Unsupported keepalive options. If you see this message "
+                    "it mean that you passed an unsupported object within "
```
jparise, line 351:
```suggestion
                    "it means that you passed an unsupported object within "
```

jparise (approved):

PR #333, opened 2021-06-30
Ensure us to close socket on MemcacheUnexpectedCloseError

4383:
During network maintenance operations or during server reboot users could
face MemcacheUnexpectedCloseError.

This kind of exception will be raised down the stack. It's a design
choice [1]. Users should manage this kind of exception.

However, at some points the opened sockets remains more or less opened
after this kind of fail on the client side [2]. It will require n+1
calls to see them recreated when the connection will be possible. That's
could be problem with services that implement several cache logics in parallel
and where their clients are based on pymemcache.

These changes simply ensure to close the socket properly. The socket
will be recreated by the client if needed.

[1] https://github.com/pinterest/pymemcache/issues/307
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1977711

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -993,7 +993,12 @@ class Client(object):
             except Exception as e:
                 raise ValueError("Unable to parse line %s: %s" % (line, e))
 
-        buf, value = _readvalue(self.sock, buf, int(size))
+        value = None
+        try:
+            buf, value = _readvalue(self.sock, buf, int(size))
+        except MemcacheUnexpectedCloseError:
+            self.close()
+            raise
         key = remapped_keys[key]
         value = self.serde.deserialize(key, value, int(flags))
 
@@ -1019,9 +1024,14 @@ class Client(object):
             self.sock.sendall(cmd)
 
             buf = b''
+            line = None
             result = {}
             while True:
-                buf, line = _readline(self.sock, buf)
+                try:
+                    buf, line = _readline(self.sock, buf)
+                except MemcacheUnexpectedCloseError:
+                    self.close()
+                    raise
                 self._raise_errors(line, name)
                 if line == b'END' or line == b'OK':
                     return result
@@ -1091,8 +1101,13 @@ class Client(object):
 
             results = {}
             buf = b''
+            line = None
             for key in keys:
-                buf, line = _readline(self.sock, buf)
+                try:
+                    buf, line = _readline(self.sock, buf)
+                except MemcacheUnexpectedCloseError:
+                    self.close()
+                    raise
                 self._raise_errors(line, name)
 
                 if line in VALID_STORE_RESULTS[name]:
@@ -1116,8 +1131,13 @@ class Client(object):
 
             results = []
             buf = b''
+            line = None
             for cmd in cmds:
-                buf, line = _readline(self.sock, buf)
+                try:
+                    buf, line = _readline(self.sock, buf)
+                except MemcacheUnexpectedCloseError:
+                    self.close()
+                    raise
                 self._raise_errors(line, cmd_name)
                 results.append(line)
             return results
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -43,6 +43,13 @@ SOCKET_KEEPALIVE_SUPPORTED_SYSTEM = {
     'Linux',
 }
 
+STORE_RESULTS_VALUE = {
+    b'STORED': True,
+    b'NOT_STORED': False,
+    b'NOT_FOUND':  None,
+    b'EXISTS': False
+}
+
 
 # Some of the values returned by the "stats" command
 # need mapping into native Python types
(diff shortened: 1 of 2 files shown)


jparise (commented):
Should we call `self.close()` instead of `sock.close()`? That wrapper handles exceptions and clears `self.sock`.

To do that, we'll need to do a little more work because these functions (`_readline` and `_readvalue`) aren't class members. It seems like the thing to do is to catch `MemcacheUnexpectedCloseError` one level up, call `self.close()`, and then re-raise the exception.

4383:
> Should we call `self.close()` instead of `sock.close()`? That wrapper handles exceptions and clears `self.sock`.
> 
> To do that, we'll need to do a little more work because these functions (`_readline` and `_readvalue`) aren't class members. It seems like the thing to do is to catch `MemcacheUnexpectedCloseError` one level up, call `self.close()`, and then re-raise the exception.

I think you are right. I was thinking about something like that (catch from the client and reraise).
I'll propose a new patch set to apply this change.

Thanks for your proposition.

cgordon:
Good catch! I can't believe this bug survived all these years, and I'm a little ashamed to have failed to close the socket in the first place (especially since the contract for the class mentions this exact case!).

4383:
> Good catch! I can't believe this bug survived all these years, and I'm a little ashamed to have failed to close the socket in the first place (especially since the contract for the class mentions this exact case!).

No problem. As long as we are able to correct it, it's not a problem.
My latest patch set implement @jparise's previous comment.

4383:
Hm... the linter complain about the complexity of the method `_store_cmd`. I guess it's due to adding my `try/except`.
Is it possible to call the linter locally?

I can't find a related tox target, and locally the `flake8` test pass without error, so I suppose that the linter is another job triggered only in CI.

4383:
So, my latest patch set introduced a new commit to outsource some processing of the `_store_cmd` method to reduce its complexity.

This patch set is now composed of two separated commits to avoid mixing topics.

pymemcache/client/base.py
```diff
@@ -905,7 +912,12 @@ def _extract_value(self, expect_cas, line, buf, remapped_keys,
             except Exception as e:
                 raise ValueError("Unable to parse line %s: %s" % (line, e))
 
-        buf, value = _readvalue(self.sock, buf, int(size))
+        value = None
+        try:
+            buf, value = _readvalue(self.sock, buf, int(size))
+        except MemcacheUnexpectedCloseError as e:
+            self.close()
+            raise e
```
jparise, line 920: I think we just want to `raise` here (without capturing and re-raising `e`). Below, too.

pymemcache/client/base.py
```diff
@@ -38,6 +38,13 @@
     b'cas':     (b'STORED', b'EXISTS', b'NOT_FOUND'),
 }
 
+STORE_RESULTS_BINDING = {
```
jparise, line 41: Let's give this a slightly different name:

```suggestion
STORE_RESULTS_VALUE = {
```

pymemcache/client/base.py
```diff
@@ -1003,19 +1020,17 @@ def _store_cmd(self, name, values, expire, noreply, flags=None, cas=None):
...
 
                 if line in VALID_STORE_RESULTS[name]:
-                    if line == b'STORED':
-                        results[key] = True
-                    if line == b'NOT_STORED':
-                        results[key] = False
-                    if line == b'NOT_FOUND':
-                        results[key] = None
-                    if line == b'EXISTS':
-                        results[key] = False
```
jparise, line 1018: The last three of these should have been `elif` instead of `if`. I don't think fixing that would have reduce the linter's complexity score though?

I'm actually equally in favor of dropping that lint rule.

jparise (commented):
Looks good! Could you also add some tests that verify that `self.sock is None` after these cases raise `MemcacheUnexpectedCloseError`?

4383:
Comments addressed and changes rebased on the top of b289c87

4383:
> `MemcacheUnexpectedCloseError`

Do we have tests that already simulate a `MemcacheUnexpectedCloseError`?
The only one that I found is related to the `shutdown` command and this one is a bit different because this exception is really expected and AFAIK the client will be closed properly.

I think we need to implement tests that simulate a `raise` of `MemcacheUnexpectedCloseError` during the call of `_readvalue` and `_readline`.

4383:
> Looks good! Could you also add some tests that verify that `self.sock is None` after these cases raise `MemcacheUnexpectedCloseError`?

Done

jparise (approved):

4383:
Thank you @jparise 
Can I ask you a new release to land these changes and the previous keepalive feature too.

jparise:
> Can I ask you a new release to land these changes and the previous keepalive feature too.

I'm watching to see how #324. That way, I can bundle all of these new features into a minor version release.

4383:
Ok no problem.

PR #334, opened 2021-07-01
Rely on tox for CI linter

4383:
These changes allow us to rely on tox when we execute our linters on CI.

They centralize the way we define our jobs and targets to use them
in standardized way locally and during CI pipelines.

With these changes we are sure to execute the same commands everywhere.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,13 @@
 [tox]
-envlist = py27, py34, py35, py36, py37, py38, py39, pypy, pypy3, py27-flake8, py39-flake8, integration
+envlist = py27, py34, py35, py36, py37, py38, py39, pypy, pypy3, flake8, integration
 skip_missing_interpreters = True
+# Automatic envs (pyXX) will only use the python version appropriate to that
+# env and ignore basepython inherited from [testenv] if we set
+# ignore_basepython_conflict.
+ignore_basepython_conflict = True
 
 [testenv]
+basepython = python3
 setenv =
     PYTHONPATH = {toxinidir}
 deps = -r{toxinidir}/test-requirements.txt
@@ -15,16 +20,11 @@ commands =
     pip install -e .
     py.test {posargs:pymemcache/test/ -m integration}
 
-[testenv:py27-flake8]
+[testenv:flake8]
+# Avoid pulling all the base requirements only to run flake8
+deps = flake8
 commands =
-    pip install flake8
-    flake8 pymemcache/
-    python setup.py check --restructuredtext
-
-[testenv:py39-flake8]
-commands =
-    pip install flake8
-    flake8 pymemcache/
+    flake8
     python setup.py check --restructuredtext
 
 [testenv:docs]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,8 @@ jobs:
     - name: Lint
       if: matrix.python-version == '2.7' || matrix.python-version == '3.9'
       run: |
-        pip install flake8
-        flake8
-        python setup.py check --restructuredtext
+        pip install tox tox-gh-actions
+        tox -e flake8
     - name: Disable IPv6 localhost
       run: |
         sudo sed -i '/::1/d' /etc/hosts


4383:
My goal is to avoid similar the problem that I faced with:

- https://github.com/pinterest/pymemcache/commit/7377e23b64ea4ff35141917cdf72d161c79657e0
- https://github.com/pinterest/pymemcache/runs/2961411929

Also my goal is to allow us to get the complexity of the code even when we execute flake8 locally.

tox.ini
```diff
@@ -15,18 +15,28 @@ commands =
     pip install -e .
     py.test {posargs:pymemcache/test/ -m integration}
 
-[testenv:py27-flake8]
+[testenv:py2.7-flake8]
```
4383, line 18: I added a dot (`.`) to allow us to rely on `matrix.python-version`

tox.ini
```diff
@@ -15,18 +15,28 @@ commands =
...
 commands =
     pip install flake8
-    flake8 pymemcache/
+    flake8 --max-complexity=18 pymemcache/
     python setup.py check --restructuredtext
 
-[testenv:py39-flake8]
+[testenv:py3.9-flake8]
```
4383, line 24: Same thing here and below

4383:
You can observe the result directly in the logs of the pipeline related to ` 339c9bd`

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, py34, py35, py36, py37, py38, py39, pypy, pypy3, py27-flake8, py39-flake8, integration
+envlist = py27, py34, py35, py36, py37, py38, py39, pypy, pypy3, py2.7-flake8, py3.9-flake8, integration
```
4383, line 2: Do we want to add `py*-codespell` here too?

4383:
These changes will avoid back and forth during pull request due to pipelines results that aren't identical to local testing.

tox.ini
```diff
@@ -15,18 +15,28 @@ commands =
...
 
-[testenv:py39-flake8]
+[testenv:py3.9-flake8]
 commands =
     pip install flake8
-    flake8 pymemcache/
+    flake8 --max-complexity=18 pymemcache/
     python setup.py check --restructuredtext
 
+[testenv:py2.7-codespell]
```
4383, line 30: Do we want `codespell` as a dedicated target or do we want it as a sub command of the `flake8` target?

It's up to you to decide. I'm ok with the both versions.

jparise (commented):
I just removed codespell and the flake8 complexity check in #335.

4383:
> I just removed codespell and the flake8 complexity check in #335.

Ah ok, so the `Lint` target of `.github/workflows/ci.yml` is more or less obsolete now, isn't?
It only check the format of the setup file, so, do we still need a target only for that?

jparise:
> > I just removed codespell and the flake8 complexity check in #335.
> 
> Ah ok, so the `Lint` target of `.github/workflows/ci.yml` is more or less obsolete now, isn't?
> It only check the format of the setup file, so, do we still need a target only for that?

It still runs the rest of the flake8 checks.

4383:
Oh yes sorry I misread your last changes, then I'll update this patch to rely on your latest changes

4383:
So I removed the `codespell` stuffs, and I kept the `flake8` transition into `tox` to centralize and uniformizing the use of it through the project.

tox.ini
jparise: I seems like there's a way to map these via a `[gh-actions]` section in `tox.ini`: https://github.com/ymyzk/tox-gh-actions#factor-conditional-settings-python-version

4383:
I think that we don't really need to define two separated `flake8` `tox` targets.

We can rely on the same target as the only difference was the interpreter in use. The `flake8` checks will remain the same whatever the python version in use, so if the code isn't valid the checks will still fail in the same way.

From this postulate we can rely on the `baseinterpreter`. Locally it will use `python3`. Still locally it will use the `python3` symlink. It will use either `3.4`, `3.6`, `3.7`, `3.8`, `3.9`, that mean that locally we are less strict and we are more compatible with the developers their environments.

However, on CI side we are more strict, we run the `Lint` job and the `flake8` check against python 2.7 and 3.9 as already did previously. The used `tox` target remain the same. We will ignore the conflict with the `basepython` and we will use the available interpreter.

jparise (approved):
There's a small benefit to running flake8 for both Python 2 and 3: it verifies its rules against each Python version's AST (which very).

That being said, we'll drop Python 2 support in the next major version, so I think it's fine to make this change now.

jparise:
> There's a small benefit to running flake8 for both Python 2 and 3: it verifies its rules against each Python version's AST (which very).
> 
> That being said, we'll drop Python 2 support in the next major version, so I think it's fine to make this change now.

I reread your diff and see that we will _still_ run flake8 for both versions, so even better. 🙂

4383:
> > There's a small benefit to running flake8 for both Python 2 and 3: it verifies its rules against each Python version's AST (which very).
> > That being said, we'll drop Python 2 support in the next major version, so I think it's fine to make this change now.
> 
> I reread your diff and see that we will _still_ run flake8 for both versions, so even better. slightly_smiling_face

Yes we still run the flake8 tests against the both version, the main difference is that we only rely on the one specific and centralized tox target.

4383:
Thanks

PR #336, opened 2021-07-02
adding code coverage

4383:
Trivial changes to add code coverage.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,10 @@ commands =
     flake8 pymemcache/
     python setup.py check --restructuredtext
 
+[testenv:coverage]
+commands =
+    py.test --cov=pymemcache
+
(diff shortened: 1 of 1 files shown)


jparise (commented):
Because we're using pytest, should we use this plugin?

https://pypi.org/project/pytest-cov/

4383:
That's not an obligation to use this plugin but that could be an option. As you want. It work with the current version, and it will work with the other plugin too.

jparise:
Let's use pytest-cov because it appears to offer better pytest integration with regard to `sys.path` handling. It also supports subproccesses, in case we need that later.

4383:
Make sense. Done with the latest patch set.

jparise (approved):

PR #337, opened 2021-07-02
[py27] Fixing a bug with bad conditions

4383:
These changes fix a condition for python 2.7 only.

Python 3 and python 2.7 behave differently during our key handling.

With python 3, in the case where the key is empty and where the key is not a
bytes object the concatenation [1] will fail and the execution will be stoped
with the raise of a TypeError exception.

With python 2 the concatenation is not a problem and so the execution
isn't stoped and that lead us to a problematic condition few line below [2].
Here we execute a condition with a boolean logic and an `or` operator.
That mean that if the first condition doesn't return `True` we will
execute the second condition, in this case we will compare an index of a
previous split with our original key. The problem here is that if our
key is empty then we will try to retrieve the index `0` which will lead
us to an IndexError (empty list).

This error is allowed with python 2.7 because the previous concatenation
doesn't fail. The execution can continue.

These changes ensure that the list is not empty before trying to
retrieve an index. This is possible by using an `and` operator in the
second logic expression. This one will stop the execution and will
ignore the index part if the list is empty.

These changes also add few unittest to check more directly this
function.

The empty key lead to a specific concatenation behaviour which
depends-on the version of python used and lead us to the part bug with
the condition.

[1] https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L110
[2] https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L116

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -113,7 +113,7 @@ def check_key_helper(key, allow_unicode_keys, key_prefix=b''):
     if len(key) > 250:
         raise MemcacheIllegalInputError("Key is too long: %r" % key)
     # second statement catches leading or trailing whitespace
-    elif len(parts) > 1 or parts[0] != key:
+    elif len(parts) > 1 or (parts and parts[0] != key):
         raise MemcacheIllegalInputError("Key contains whitespace: %r" % key)
     elif b'\00' in key:
         raise MemcacheIllegalInputError("Key contains null: %r" % key)
--- a/pymemcache/test/test_client.py
+++ b/pymemcache/test/test_client.py
@@ -55,7 +55,7 @@ def is_ipv6(address):
 
 
 @pytest.mark.parametrize(
-    'key, allow_unicode_keys,key_prefix,ex_exception,ex_excinfo,ignore_py27',
+    'key,allow_unicode_keys,key_prefix,ex_exception,ex_excinfo,ignore_py27',
     [
         (u'b'*251, True, b'',
          MemcacheIllegalInputError, 'Key is too long', False),
--- a/pymemcache/test/test_client.py
+++ b/pymemcache/test/test_client.py
@@ -24,6 +24,7 @@ import mock
 import platform
 import re
 import socket
+import sys
 import unittest
 
 import pytest
@@ -32,7 +33,8 @@ from pymemcache.client.base import (
     PooledClient,
     Client,
     normalize_server_spec,
-    KeepaliveOpts
+    KeepaliveOpts,
+    check_key_helper
 )
 from pymemcache.exceptions import (
     MemcacheClientError,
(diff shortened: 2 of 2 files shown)


4383:
My unit tests doesn't really reflect this bug but if you want I can implement conditional tests that will reflect the problem.
I think that the problem is enough explicit to be understood by reading the condition.
The tricky part is more the concatenation part of the issue, that isn't evident at first glance.
Up to you.

4383:
I added tests with d4930ed to highlight the difference between python 2 and python 3.

4383:
Another way to observe that is to execute:

```
$ python2.7 -c "print('' + b'')"

$ python3.7 -c "print('' + b'')"
  File "<string>", line 1, in <module>
TypeError: can only concatenate str (not "bytes") to str
$ python3.6 -c "print('' + b'')"
  File "<string>", line 1, in <module>
TypeError: must be str, not bytes
$ python3.5 -c "print('' + b'')"
  File "<string>", line 1, in <module>
TypeError: Can't convert 'bytes' object to str implicitly
```

pymemcache/test/test_client.py
```diff
@@ -52,6 +54,74 @@ def is_ipv6(address):
...
+        {
+            'key': None,
+            'allow_unicode_keys': True,
+            'key_prefix': b'',
+            'expected_exception': TypeError,
+            'expected_excinfo': None
+        },
+        {
+            # This test won't fail with python 2.7
+            'key': b"",
```
4383, line 90: Without the fix above, with python 2.7, this test will lead to an `IndexError`.

pymemcache/client/base.py
```diff
@@ -113,7 +113,7 @@ def check_key_helper(key, allow_unicode_keys, key_prefix=b''):
     if len(key) > 250:
         raise MemcacheIllegalInputError("Key is too long: %r" % key)
     # second statement catches leading or trailing whitespace
-    elif len(parts) > 1 or parts[0] != key:
+    elif len(parts) > 1 or (len(parts) > 0 and parts[0] != key):
```
jparise, line 116:
```suggestion
    elif len(parts) > 1 or (parts and parts[0] != key):
```

pymemcache/test/test_client.py
```diff
@@ -52,6 +54,74 @@ def is_ipv6(address):
     return re.match(r'^[0-9a-f:]+$', address)
 
 
+@pytest.mark.unit()
+def test_check_key_helper():
+    testing_matrix = (
```
jparise, line 59: This would be a good place to use [`pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html#parametrize).

4383:
Done

pymemcache/test/test_client.py
```diff
@@ -52,6 +54,45 @@ def is_ipv6(address):
     return re.match(r'^[0-9a-f:]+$', address)
 
 
+@pytest.mark.parametrize(
+    'key, allow_unicode_keys,key_prefix,ex_exception,ex_excinfo,ignore_py27',
```
jparise, line 58:
```suggestion
    'key,allow_unicode_keys,key_prefix,ex_exception,ex_excinfo,ignore_py27',
```

pymemcache/test/test_client.py
```diff
@@ -52,6 +54,45 @@ def is_ipv6(address):
...
+        (b"", False, '', TypeError, None, True),
+    ])
+@pytest.mark.unit()
+def test_check_key_helper_failing_conditions(key, allow_unicode_keys,
+                                             key_prefix, ex_exception,
+                                             ex_excinfo, ignore_py27):
+
+    if ignore_py27:
+        if sys.version_info < (3, 0, 0):
+            return
```
jparise, line 77:
```suggestion
    if ignore_py27 and sys.version_info < (3, 0, 0):
        pytest.skip("skipping for Python 2.7")
```

jparise (approved):

PR #339, opened 2021-07-05
Fix doc rendering

4383:
The KeepaliveOpts's taken parameters aren't well renderer in the doc.
Their docstrings are missing.

This patch fix that.


4383:
Example given without these changes:
https://pymemcache.readthedocs.io/en/latest/apidoc/pymemcache.client.base.html#pymemcache.client.base.KeepaliveOpts

jparise (approved):

PR #343, opened 2021-07-15
Adding copyright and licence missing headers

4383:
Uniformizing source code licencing to explicitly highlight legal aspects on source code.

### DIFF
--- a/pymemcache/__init__.py
+++ b/pymemcache/__init__.py
@@ -1,3 +1,16 @@
+# Copyright 2021 Pinterest.com
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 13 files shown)


jparise:
This is a little more nuanced than just adding copyright headers to every module. I believe adding the license block is fine (although not required given the top-level `LICENSE`), but retroactively adding a `Copyright` statement should only be done by the files' original (majority?) author(s).

4383:
> This is a little more nuanced than just adding copyright headers to every module. I believe adding the license block is fine (although not required given the top-level `LICENSE`), but retroactively adding a `Copyright` statement should only be done by the files' original (majority?) author(s).

The goal of these changes is mostly to add the licence header, I proposed the copyright part mostly to standardize all the files.

Concerning the copyright part I think that you're right, it is not needed to add them retroactively.

I'll drop the copyright parts.

jparise:
I'm going to close this for now. Please re-open if you have time to revisit this.

PR #345, opened 2021-08-10
Use sphinxcontrib-apidoc extension for readthedocs builds

JWCook:
This PR replaces the `run_apidoc` event handler with the [sphinxcontrib-apidoc](https://github.com/sphinx-contrib/apidoc) extension. It does pretty much the same thing, but it's a little more convenient. Feel free to close this PR if you don't find that useful, but personally it's saved me quite a few headaches.

Notes:
* `apidoc_separate_modules` is equivalent to the `-e` option
* `apidoc_module_dir` and `apidoc_output_dir` are relative to the documentation source dir (`/docs`)
* I updated `tox.ini` so local Sphinx builds will use the apidoc config in `conf.py` instead of separately running the `sphinx-apidoc` command
* I added a `.readthedocs.yml` to install `docs-requirements.txt`. If you already have that configured in your settings on readthedocs.org, that file can be removed.

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,3 +1,4 @@
 sphinx==3.0.3
 sphinx_rtd_theme
+sphinxcontrib-apidoc
 sphinxcontrib-napoleon
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -36,6 +36,7 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.napoleon',
     'sphinx.ext.ifconfig',
+    'sphinxcontrib.apidoc',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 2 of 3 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,11 @@
+# Read the Docs configuration file
+version: 2
+
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+
+python:
+  version: 3.8
```
jparise, line 9: Any reason not to use 3.9?

jparise:
> * I added a `.readthedocs.yml` to install `docs-requirements.txt`. If you already have that configured in your settings on readthedocs.org, that file can be removed.

Yes, we already have this configured on readthedocs.org. I think we don't need the additional configuration file.

JWCook:
> > * I added a `.readthedocs.yml` to install `docs-requirements.txt`. If you already have that configured in your settings on readthedocs.org, that file can be removed.
> 
> Yes, we already have this configured on readthedocs.org. I think we don't need the additional configuration file.

Okay, I removed it.

.readthedocs.yml
JWCook: Nope! rtd was a little late to support python 3.9, but it looks like it's supported now: https://github.com/readthedocs/readthedocs.org/issues/7554

jparise (approved):
This looks good. Thanks for the improvement!

JWCook:
You're very welcome. Thanks for the great library!

PR #347, opened 2021-09-08
Update README.rst - (broken link)

juandoleal:
Hey guys,

Proposing to replace the following broken link: http://code.mixpanel.com/2012/07/16/we-went-down-so-we-wrote-a-better-pure-python-memcache-client/

This is the updated link: https://engineering.mixpanel.com/we-went-down-so-we-wrote-a-better-pure-python-memcache-client-b409a9fe07a9

Best,
Juan


jparise (approved):
Thank you!

Issue #350, opened 2021-09-21
ignore_exc should return default values


Justylol:
Its better for ignore_exc to use default value on return, instead of None. It will be real missing cache behaviour and will help with errors when default expected. For example: 

Django==3.2
djangorestframework==3.11.1
pymemcache==3.5.0

 rest_framework.throttling module
 
 
```
        self.history = self.cache.get(self.key, [])
        self.now = self.timer()

        # Drop any requests from the history which have now passed the
        # throttle duration
        while self.history and self.history[-1] <= self.now - self.duration:
            self.history.pop()
        if len(self.history) >= self.num_requests:
            return self.throttle_failure()
```
            
This code with in Django with PyMemcacheCache cache backend and ignore_exc option leads to an error  `object of type 'NoneType' has no len()` because instead of returning list as missing cache result, it returns None.

jogo:
Thank you for the bug report, #351 should address this

PR #354, opened 2021-10-07
Return default from hash client when using positional argument

Pankrat:
When using `HashClient` with `ignore_exc`, `get` would always return
`None` if no server is available and the default is passed as a
positional argument. The other clients return the default
value in this case. An earlier fix only had the desired effect when
passing `default` as a keyword argument.

For example, [Django passes the default as a positional argument](https://github.com/django/django/blob/main/django/core/cache/backends/memcached.py#L72).

Return the default value so `HashClient` behaves like the other
clients even when using

Fixes another variation of issue #350

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Unreleased
+----------
+* ``Client.get`` returns the default when using ``ignore_exc`` and if memcached
+  is unavailable
+
 New in version 3.5.0
 --------------------
 * Sockets are now closed on ``MemcacheUnexpectedCloseError``.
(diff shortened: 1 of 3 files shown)


Pankrat:
This should probably be changed to "Unreleased". I just put the version
from `pymemcache/__init__.py` there as it looked like this is going to be
the upcoming version.

Joe Gordon ***@***.***> schrieb am Do., 7. Okt. 2021, 22:02:

bitcity:
@Pankrat Just curious if it's better to keep the function parameters (declaration) consistent and have something like : 

```
    def get(self, key, *args, **kwargs):
        if 'default' in kwargs:
            default = kwargs.get('default')
        elif args:
            default = args[0]
        else:
            default = None
        return self._run_cmd('get', key, default, *args, **kwargs)

```
as opposed to 

```
    def get(self, key, default=None, **kwargs):
```

PR #358, opened 2021-12-03
Black

jogo:
Apply black https://black.readthedocs.io/en/stable/

See individual commit messages for more details of each change


jparise (commented):
I think we want to add the the black CI action to the existing `ci.yml` workflow (or move those lint steps out and into a generic `lint.yml` workflow).

setup.cfg
```diff
@@ -44,3 +44,4 @@ markers =
 show-source = True
 max-line-length = 80
```
jparise, line 45: I think we should also set this to `88`: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

.github/workflows/ci.yml
```diff
@@ -50,3 +50,11 @@ jobs:
...
         ports:
         - 11211/tcp
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+        with:
+          version: "21.5b1"
```
jparise, line 60: Given that black is already configured in `tox.ini`, you could drop this step entirely and just update the existing lint step above:

```diff
@@ -29,7 +29,7 @@ jobs:
       if: matrix.python-version == '3.9'
       run: |
         pip install tox tox-gh-actions
-        tox -e flake8
+        tox -e black,flake8
     - name: Disable IPv6 localhost
       run: |
         sudo sed -i '/::1/d' /etc/hosts
```

jparise (commented):
I think we could also add a `pyproject.toml` configuration that explicitly sets the list of target versions (because pymemcache only supports Python 3.6+):

```toml
[tool.black]
target-version = ['py36', 'py37', 'py38', 'py39']
```

If we wanted  small migration diff, we could also add `skip-string-normalization = true`, but I think it's fine to get that out of the way here.

jogo:
Confirmed black is running correctly https://github.com/pinterest/pymemcache/runs/4411097331?check_suite_focus=true

jparise (approved):
Nice work!

PR #373, opened 2022-02-15
Initial support for mypy-based type checking

jparise:


cgordon:
Is the idea that you will eventually add some static type checking to `pymemcache`? I'm intrigued.

jparise:
> Is the idea that you will eventually add some static type checking to `pymemcache`? I'm intrigued.

That's right, with a later goal of trying out [mypyc](https://github.com/mypyc/mypyc) to produce a binary (compiled) artifact for benchmarking.

cgordon (approved):

PR #374, opened 2022-02-15
Remove tox from the testing pipeline

jparise:
tox has become somewhat redundant given the capabilities of pytest and
GitHub Actions, and removing it means one less tool to maintain and
configure.


jparise:
Dropping in favor of #375

Issue #359, opened 2022-01-14
HashClient failover - How a node is automatically rebalanced when a node of cluster goes down?


4383:
Hello everyone,

We've some questions regarding how the HashClient's failover mechanisms works and what we should expect when we use memcached cluster with pymemcache.

Here is an example of scenario.

We have 3 cluster nodes:

```
192.168.0.3:11211
192.168.0.4:11211
192.168.0.4:11211
```

We instantiate a HashClient by passing those nodes to the `servers` param.

This client is alive and the socket remains connected and live in an execution loop, we call `set` and `get` a couple of time, the client will communicate with one of these 3 servers accordingly to the the hash. While this server is up the client will use only this server. However at some point we kill all the 3 nodes and we restart only two nodes that aren't the server used by the client initialy, what do we should expect at this point?

Does the client will automatically switch the connection (by establishing a new socket) to one of these 2 servers and restart to set and get things from it?

Apparently, if we understand correctly your doc, [it said that yes, the server will be automatically rebalanced depending on if a server goes down](https://pymemcache.readthedocs.io/en/latest/getting_started.html#using-a-memcached-cluster).

Here is a sample of code that more or less summarize the previously described scenario:

https://github.com/4383/oslo.cache-labs/blob/master/labs/oslo.cache/pymemcache_backend/app.py
https://github.com/4383/oslo.cache-labs/blob/master/labs/oslo.cache/pymemcache_backend/lab/app.conf

This snippet of code rely on the stack [oslo.cache](https://review.opendev.org/c/openstack/oslo.cache/+/803747) / [dogpile.cache](https://github.com/sqlalchemy/dogpile.cache/commit/36390997c6350a606c7cabe4ea85dd05f4bddd52) / pymemcache

Our problem is that with this scenario we didn't observed the server switch within our client (either in this snippet or even in a more serious testing in openstack), so either we misundertood something or we missed something during our tests.

We'd expect to see the client mark this server as dead and switch the connection to another server that is up.
That's not what we observed.

Please can you give us more details about what we should expect with this scenario.

Thanks in advance.

martinnj:
I think you might've misunderstood what `HashClient` does. 

`HashClient` will always try to distribute keys evenly on all given nodes using the `hasher` provided in the constructor. By default this is the built-in [Rendezvous hasher](https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/rendezvous.py).

[...]

PR #381, opened 2022-03-11
Add more documentation on HashClient to getting started.

martinnj:
By request from @jparise in #359.

I just wrote about the same content and style I did in the issue. But If you guys would rather have the format changed. Let me know.

I also took the liberty of adding my name to credits in the README. Hope that is okay.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -85,7 +85,7 @@ used to change how keys are distributed. Changing the hashing algorithm can be
 done by setting the ``hash_function`` argument in the ``RendezvousHash``
 constructor.
 
-Rebalancing in the :class:`pymemcache.client.hash.HashClient` functions as'
+Rebalancing in the :class:`pymemcache.client.hash.HashClient` functions as
 follows:
 
 1. A :class:`pymemcache.client.hash.HashClient` is created with 3 nodes,
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -77,6 +77,47 @@ on if a server goes down.
     client.set('some_key', 'some value')
     result = client.get('some_key')
 
+Key distribution is handled by the ``hasher`` argument in the constructor. The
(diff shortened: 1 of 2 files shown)


docs/getting_started.rst
```diff
@@ -77,6 +77,47 @@ on if a server goes down.
...
 
+Key distribution is handled by the ``hasher`` argument in the constructor. The
+default is the built-in :class:`pymemcache.client.rendezvous.RendezvousHash`
+hasher. It uses the built-in :class:`pymemcache.client.murmur3.murmur3_32`
+implementation to distribute keys on servers. Overriding these two parts can be
+used to change how keys are distributed. Changing the hashing algorithm can be
+done by setting the ``hash_function`` argument in the ``RendezvousHash``
+constructor.
+
+Rebalancing in the :class:`pymemcache.client.hash.HashClient` functions as'
```
jparise, line 88: Stray apostrophe?

```suggestion
Rebalancing in the :class:`pymemcache.client.hash.HashClient` functions as
```

jparise (approved):
Excellent!

docs/getting_started.rst
martinnj: Oh yeah, my bad. Thought I proofread it. Good catch. :+1:

glory-123: Hi @martinnj ,
The hashclient tries to ``get("key2")`` but detects the node as down. This causes it to mark the node as down. Removing it from the hasher. The hasclient can attempt to retry the operation based on the ``retry_attempts`` and ``retry_timeout`` arguments.
In the above point will it  retry the operation or client?

martinnj: @glory-123 I don't think this PR is the place for that question, but the docstring of the class should hgopefully clarify the behavior in question:
https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/hash.py#L66-L71
It will attempt the operation `retry_attempts` times with `retry_timeout` time in between before marking the node as dead for `dead_timeout` seconds.

PR #388, opened 2022-03-25
Handle a blank stat value

liquidpele:
A stat with a blank value makes the .stats() call fail with index error.  

```
memcache_connection.stats() 
File "/lib/python3.6/site-packages/pymemcache/client/base.py", line 741, in stats 
result = self._fetch_cmd(b'stats', args, False) 
File "/lib/python3.6/site-packages/pymemcache/client/base.py", line 921, in _fetch_cmd 
result[key_value[1]] = key_value[2]
IndexError: list index out of range
```

The line in question we saw that caused this:

```
b'STAT ep_initfile '
```

This fixes things to no longer blow up.   I chose to use a blank binary string instead of None because I suspect other code might make assumptions about the data being binary strings due to this bug.

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -1038,7 +1038,7 @@ class Client:
                     result[key] = value
                 elif name == b"stats" and line.startswith(b"STAT"):
                     key_value = line.split()
-                    result[key_value[1]] = key_value[2] if len(key_value) > 2 else b''
+                    result[key_value[1]] = key_value[2] if len(key_value) > 2 else b""
                 elif name == b"stats" and line.startswith(b"ITEM"):
                     # For 'stats cachedump' commands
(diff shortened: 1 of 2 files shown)


jparise (commented):
Would you be able to also add a unit test that covers this case?

jparise (approved):
Looks great. Thank you!

liquidpele:
Awesome, thanks for the fast responses.   Any chance for a new release soon that will include this so I can import it into our local company repo from pypi?

jparise:
> Awesome, thanks for the fast responses. Any chance for a new release soon that will include this so I can import it into our local company repo from pypi?

[v3.5.2](https://github.com/pinterest/pymemcache/releases/tag/v3.5.2) has been tagged and released on PyPI.

PR #402, opened 2022-06-07
Small buffer pass optimization as discussed in #395.

martinnj:
Basically ensure the client only does one pass over the buffer instead of two.
Exact thread: https://github.com/pinterest/pymemcache/pull/395#discussion_r890288417


jparise (approved):

jparise:
Thank you!

PR #414, opened 2022-08-17
Fixed Typos in Docs

10kartik:
Fixed a few grammatical and sentence construction mistakes from docs.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -275,7 +275,7 @@ You can instantiate all the classes and clients offered by pymemcache.
 
 Your client will remain open until you decide to close it or until you decide
 to quit your interpreter. It can allow you to see what happens if your server
-is abruptly closed. Below is a by-example.
+is abruptly closed. Below is an example.
 
 Starting your server:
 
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -274,8 +274,8 @@ pymemcache and fetch all the needed requirements at run. See the example below:
 You can instantiate all the classes and clients offered by pymemcache.
 
 Your client will remain open until you decide to close it or until you decide
-to quit your interpreter. It can allow you to see what's happen if your server
-is abruptly closed. Below is an by example.
+to quit your interpreter. It can allow you to see what happens if your server
+is abruptly closed. Below is a by-example.
```
jogo, line 278: Out of curiosity can you point me to the grammar guideline for the dash in 'by-example' I haven't personally seen that one before.

jogo (approved):
Thank you for the cleanup! I have one very tiny question out of curiosity, will merge afterwards.

docs/getting_started.rst
jparise, line 278: I think this would be best as:

```suggestion
is abruptly closed. Below is an example.
```

10kartik: yeah, sorry! There shouldn't have been '-' but 'by example' instead or 'an example' as suggested by @jparise below.

PR #415, opened 2022-08-17
Start to add type hints

jogo:
First pass at adding some type hints to pymemcache to make it easier to
develop against etc.


pymemcache/client/base.py
```diff
@@ -53,23 +51,24 @@
 }
 
 ServerSpec = Union[Tuple[str, int], str]
+KeyInputSpec = Union[bytes, str]
```
jparise, line 54: Maybe just `Key`, as this is the only public key type?

pymemcache/client/base.py
```diff
@@ -439,7 +440,14 @@ def close(self):
 
     disconnect_all = close
 
-    def set(self, key, value, expire=0, noreply=None, flags=None):
+    def set(
+        self,
+        key: KeyInputSpec,
+        value: Any,
```
jparise, line 446: I'd like to have some constraints on `value`, but that might need to wait until we have more typing machinery in place for the serde interface.

pymemcache/client/base.py
```diff
@@ -608,7 +629,7 @@ def cas(self, key, value, cas, expire=0, noreply=False, flags=None):
             b"cas", {key: value}, expire, noreply, flags=flags, cas=cas
         )[key]
 
-    def get(self, key, default=None):
+    def get(self, key, default: Optional[Any] = None):
```
jparise, line 632: `key` type?

pymemcache/client/base.py
```diff
@@ -698,7 +719,9 @@ def delete(self, key, noreply=None):
             return True
         return results[0] == b"DELETED"
 
-    def delete_many(self, keys, noreply=None):
+    def delete_many(
+        self, keys: List[KeyInputSpec], noreply: Optional[bool] = None
```
jparise, line 723: Does `keys` _need_ to be a list? It looks like we can handle any iterable.

pymemcache/client/base.py
```diff
@@ -275,15 +276,15 @@ def __init__(
...
-        ignore_exc=False,
+        timeout: Optional[float] = None,
+        no_delay: bool = False,
+        ignore_exc: bool = False,
         socket_module=socket,
-        socket_keepalive=None,
-        key_prefix=b"",
+        socket_keepalive: Optional[KeepaliveOpts] = None,
+        key_prefix: bytes = b"",
         default_noreply=True,
```
jparise, line 285: I think this is a `bool`?

pymemcache/client/base.py
jogo: agreed

jparise (approved):

PR #426, opened 2022-09-30
Add more type annotations

jogo:
Continue to improve the type annotation coverage

### DIFF
--- a/pymemcache/pool.py
+++ b/pymemcache/pool.py
@@ -14,24 +14,27 @@
 
 import collections
 import contextlib
-import sys
 import threading
 import time
+from typing import Callable, Optional, TypeVar, Deque, List, Generic, Iterator
 
 
-class ObjectPool:
+T = TypeVar("T")
+
+
+class ObjectPool(Generic[T]):
     """A pool of objects that release/creates/destroys as needed."""
 
     def __init__(
         self,
-        obj_creator,
-        after_remove=None,
-        max_size=None,
-        idle_timeout=0,
-        lock_generator=None,
+        obj_creator: Callable[[], T],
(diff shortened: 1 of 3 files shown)


pymemcache/pool.py
```diff
@@ -43,7 +46,7 @@ def __init__(
             raise ValueError('"max_size" must be a positive integer')
         self.max_size = max_size
         self.idle_timeout = idle_timeout
-        self._idle_clock = time.time if idle_timeout else int
+        self._idle_clock = time.time if idle_timeout else lambda: 0
```
jparise, line 49: Would `float` work here instead?

jparise (approved):

pymemcache/pool.py
jogo: mypy throws the following error: `pymemcache/pool.py:113: error: "object" not callable`

jparise: Ah. This might need to be:

```suggestion
        self._idle_clock: Callable[[], float] = time.time if idle_timeout else float
```

jogo: ```pymemcache/pool.py:49: error: Incompatible types in assignment (expression has type "object", variable has type "Callable[[], float]")```

```python
from typing import Callable
import time

a: Callable = int
b: Callable = time.time
c: Callable = time.time
idle_timeout = False
d: Callable = time.time if idle_timeout else float
```
```
$ mypy foo.py
foo.py:8: error: Incompatible types in assignment (expression has type "object", variable has type "Callable[..., Any]")
Found 1 error in 1 file (checked 1 source file)```

Looks like the issue is the if statement will refactor

jparise (approved):

PR #435, opened 2022-10-17
Update project urls to include link to docs

jogo:
Make it easier to find the documentation from https://pypi.org/project/pymemcache/

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls


jparise (approved):
Great!

jparise (approved):

Issue #450, opened 2022-12-06
KeyError caused by different key type.


fabiocaccamo:
Hi, thank you for this great lib!

I'm using `pymemcache` in a `Python 3.8` + `Django 3.2` project and I experienced an unexpected `KeyError` (specifically using the `django.views.decorators.cache.cache_page` decorator).

Unfortunately I'm not able to replicate it, but checking the logs the problem is clear:
**Key is of type `bytes` while it is expected to be of type `str`.**

The error has been raised here:

<img width="907" alt="Screenshot 2022-12-06 at 10 33 08" src="https://user-images.githubusercontent.com/1035294/205873888-f7023d91-41eb-4432-9a87-3a8debb1cd2b.png">

Here variable values at error time:

<img width="908" alt="Screenshot 2022-12-06 at 10 34 23" src="https://user-images.githubusercontent.com/1035294/205874120-6c47d968-b79b-4fd8-a636-bef9cf865e0c.png">

This is the error raw stack trace:
```python
KeyError: b':1:views.decorators.cache.cache_page..GET.17467d2ae029dd3e1ccc2200de93e2f6.d41d8cd98f00b204e9800998ecf8427e.en'
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "sitebase/views.py", line 95, in router
    page_response = page_obj.get_response(
  File "sitebase/models.py", line 195, in get_response
    view_response = view_function(request)
  File "django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "django/utils/decorators.py", line 122, in _wrapped_view
    result = middleware.process_request(request)
  File "django/middleware/cache.py", line 145, in process_request
    cache_key = get_cache_key(request, self.key_prefix, 'GET', cache=self.cache)
  File "django/utils/cache.py", line 362, in get_cache_key
    headerlist = cache.get(cache_key)
  File "django/core/cache/backends/memcached.py", line 77, in get
    return self._cache.get(key, default)
  File "pymemcache/client/hash.py", line 347, in get
    return self._run_cmd("get", key, default, default=default, **kwargs)
  File "pymemcache/client/hash.py", line 322, in _run_cmd
    return self._safely_run_func(client, func, default_val, *args, **kwargs)
  File "pymemcache/client/hash.py", line 211, in _safely_run_func
[...]

PR #452, opened 2022-12-07
Fix `KeyError` caused by wrong key type. #450 

fabiocaccamo:
Fix `KeyError` caused by wrong key type. #450


jogo:
thanks for the fix, wouldn't this just surprises the issue not fix it? 

mind adding a unit test and your name to the credits https://github.com/pinterest/pymemcache#credits

Issue #470, opened 2023-01-23
MockMemcacheClient return string as bytes


iurisilvio:
This test fails because it returns `b"world"` instead of a string.

```
import pytest
import pymemcache
from pymemcache.test.utils import MockMemcacheClient

@pytest.mark.unit()
def test_get_set_string():
    client = MockMemcacheClient(serde=pymemcache.serde.pickle_serde)
    assert client.get("hello") is None

    client.set("hello", "world")
    assert client.get("hello") == "world"
```

(updated this example using `serde`)

iurisilvio:
Oh this is behavior of the client too. Unexpected for me...

PR #471, opened 2023-01-23
Check if value can be encoded but preserve original value.

iurisilvio:
Ref #470.


pymemcache/test/test_utils.py
```diff
@@ -110,4 +120,4 @@ def test_prepand_append():
     client.set(b"k", "1")
     client.append(b"k", "a")
     client.prepend(b"k", "p")
-    assert client.get(b"k") == b"p1a"
+    assert client.get(b"k") == "p1a"
```
iurisilvio, line 123: I'm not sure this test is correct. Without `pickle_serde`, it really must be bytes.

pymemcache/test/test_utils.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+import pymemcache
```
jogo, line 3: how about a more specific import `from pymemcache.serde import pickle_serde`

pymemcache/test/test_utils.py
```diff
@@ -12,6 +13,15 @@ def test_get_set():
     assert client.get(b"hello") == 12
 
 
+@pytest.mark.unit()
+def test_get_set_string():
+    client = MockMemcacheClient(serde=pymemcache.serde.pickle_serde)
+    assert client.get("hello") is None
+
+    client.set("hello", "world")
```
jogo, line 21: what about using a variable here?

```
value = "world"
client.set("hello", value)
...
```

pymemcache/test/test_utils.py
```diff
@@ -12,6 +13,15 @@ def test_get_set():
     assert client.get(b"hello") == 12
 
 
+@pytest.mark.unit()
+def test_get_set_string():
```
jogo, line 17: FYI there are also some tests under `pymemcache/test/test_client.py::TestMockClient`

pymemcache/test/test_utils.py
iurisilvio: Changed.

pymemcache/test/test_utils.py
iurisilvio: Done.

pymemcache/test/test_utils.py
iurisilvio: I reverted this change and implemented a fix for the right test.

pymemcache/test/utils.py
```diff
@@ -39,6 +39,14 @@ def __init__(
 
         self._contents = {}
 
+        def _serializer(key, value):
```
iurisilvio, line 42: Maybe `LegacyWrappingSerde._default_serializer` should convert string to bytes instead of this hack. On real client, this conversion happens on memcache write.

pymemcache/test/test_utils.py
```diff
@@ -12,6 +13,26 @@ def test_get_set():
...
+    client = MockMemcacheClient()
+    assert client.get("hello") is None
+
+    value = "world"
+    client.set("hello", value)
+    assert client.get("hello") == b"world"
+
+
+@pytest.mark.unit()
+def test_get_set_string_with_serde():
```
iurisilvio, line 27: This test was broken. It return a string with the real client, but mock implementation was returning bytes.

iurisilvio:
I think it is good for another review.

jogo (approved):
thanks

PR #476, opened 2023-02-02
Consistent behaviour for incr, decr for Hash and Base

### LINKED ISSUE #475: Different signature/behaviour of incr, decr for Hash and Base.
Hash implementation of incr is defined that it returns False as default in case of hashed key to assigned server failures.

```
    def incr(self, key, *args, **kwargs):
        return self._run_cmd("incr", key, False, *args, **kwargs)

    def decr(self, key, *args, **kwargs):
        return self._run_cmd("decr", key, False, *args, **kwargs)
```

However definition of incr in base.py is:
```
    def incr(
        self, key: Key, value: int, noreply: Optional[bool] = False
    ) -> Optional[int]:
```

So it would be better to return None than False by default also in Hash.

If you agree I can prepare a PR.

matejsp:
This fixes different behaviour between base and hash for incr, decr: Fixes #475

### DIFF
--- a/pymemcache/client/hash.py
+++ b/pymemcache/client/hash.py
@@ -359,10 +359,10 @@ class HashClient:
         return self._run_cmd("gats", key, default, default=default, **kwargs)
 
     def incr(self, key, *args, **kwargs):
-        return self._run_cmd("incr", key, False, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


jparise (approved):
This looks good to me. Thanks!

PR #535, opened 2023-08-16
Adding initial commit for the bloacking functionality in the get func…

JoJo10Smith:
This will add the blocking functionality to the get function with the following optional parameters:

- max_attempts: this will determine how many attempts to try and find a free object in the specified time
- max_timeout: the maximum time in seconds to try and connect after which the normal functionality of the method will occur.

### DIFF
--- a/pymemcache/pool.py
+++ b/pymemcache/pool.py
@@ -72,29 +72,40 @@ class ObjectPool(Generic[T]):
             raise
         self.release(obj)
 
-    def get(self):
-        with self._lock:
-            # Find a free object, removing any that have idled for too long.
-            now = self._idle_clock()
-            while self._free_objs:
-                obj = self._free_objs.popleft()
-                if now - obj._last_used <= self.idle_timeout:
-                    break
+    def get(self, **options):
+        #get the option parameters if the user needs blocking.
+        max_attempts = max(options.get('max_attempts',1),1)
+        max_timeout = options.get('max_timeout')
+        current_attempt = 0
 
-                if self._after_remove is not None:
-                    self._after_remove(obj)
-            else:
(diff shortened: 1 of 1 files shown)


JoJo10Smith:
I realized that there was a better implementation which I will explain below: 

- Get the `current_time` and set an `end_time` to be the `current_time` + optional `timeout`, if `timeout` is not given then set the `end_time` = `current_time`. 
- Start the function as usual but now while the `current_time` <= `end_time `keep trying to remove any free objects that have idled for too long. 
- If a free object is then return `obj`, if not then get the `current_time` again, if it is before the `end_time` try removing free objects again.
- Once the time limit is reached if there are no free objects (because the while loop has not been broken) we create one, if there are too many raise a `RuntimeError.` else return `obj`

I look forward to hearing any feedback/ responding to questions
Thanks
Jordan

pymemcache/pool.py
```diff
@@ -72,17 +72,27 @@ def get_and_release(self, destroy_on_fail=False) -> Iterator[T]:
...
             # Find a free object, removing any that have idled for too long.
-            now = self._idle_clock()
-            while self._free_objs:
-                obj = self._free_objs.popleft()
-                if now - obj._last_used <= self.idle_timeout:
-                    break
-
-                if self._after_remove is not None:
-                    self._after_remove(obj)
+            while current_time <= end_time:
```
jogo, line 83: if I understand this correctly this results in busy waiting all while holding a lock. Which seems like a behavior we don't want to introduce.

jogo (changes requested):
thanks for the PR, a few inline comments and once those are sorted out better inline documentation, unit tests and updating the changelog would be great.

PR #597, opened 2024-08-02
Update ci.yml and remove python 3.7

yurushao:
python 3.7 is no longer supported.


jparise:
Remove Python 3.7 from the list of supported versions in `setup.cfg` too?

PR #605, opened 2024-08-07
Update pyproject.toml

yurushao:
1. default mypy python version to 3.10
2. remove py37 from black target version list


jparise (approved):

jparise:
I think we can close this in favor of #620.

PR #608, opened 2024-08-13
Run tests on Python 3.13

cclauss:
* https://www.python.org/download/pre-releases
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases

Blocked by:
* python-greenlet/greenlet#396


jparise:
I think we can close this in favor of #620.

Issue #633, opened 2025-03-10
_parse_hex function is fed with base 8 instead of base 16.


redorff:
in base.py line 74 : _parse_hex function is fed with base 8 instead of base 16.

PR #634, opened 2025-03-10
_parse_hex function should declare the base as 16 not 8

redorff:
https://github.com/pinterest/pymemcache/issues/633

### DIFF
--- a/pymemcache/client/base.py
+++ b/pymemcache/client/base.py
@@ -72,7 +72,7 @@ def _parse_float(value: bytes) -> float:
 
 
 def _parse_hex(value: bytes) -> int:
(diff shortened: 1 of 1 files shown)


jparise:
It looks like some users of this function were expecting the octal behavior.

redorff:
Ok, then, shouldn't the function be called "_parse_oct" rather than "_parse_hex" ?

jparise:
> Ok, then, shouldn't the function be called "_parse_oct" rather than "_parse_hex" ?

I think so. Or there should be two functions, and the callers should use the one they need.

I also don't see a reason why the callers couldn't just use `int(...)` directly.

Issue #612, opened 2024-09-26
'NoneType' object has no attribute 'recv'


BertrandBordage:
This error happens occasionally at random in my application.
The application:
- uses `gunicorn --workers 1 --threads 8`, so with gthread
- connects to memcached through a UNIX socket
- the application and memcached run in separate Docker Compose services, with the UNIX socket being shared through a named volume

Cache configuration in Django:
```python
CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.memcached.PyMemcacheCache',
        'LOCATION': 'unix:/var/run/memcached/memcached.sock',
        'OPTIONS': {
            'use_pooling': True,  # Keeps connections alive, more reliable with multiple threads per worker.
            'default_noreply': True,  # Do not wait for a reply when we do not need it.
            'connect_timeout': 10.0,  # seconds
            'timeout': 10.0,  # seconds
        },
    },
}

```

`'ignore_exc': True` could be specified to consider this error a cache miss (if this type of error is correctly caught), but this means we would not know whether 1% or 100% of the cache calls are failing.

Versions:
pymemcache 4.0.0
memcached 1.6.31 (from the memcached:1.6.31-alpine Docker image)
gunicorn 23.0.0
Python 3.12.4
Django 5.0.7

Relevant traceback:
```python
  File "/srv/.local/lib/python3.12/site-packages/django/core/cache/backends/memcached.py", line 75, in get
    return self._cache.get(key, default)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/.local/lib/python3.12/site-packages/pymemcache/client/hash.py", line 347, in get
    return self._run_cmd("get", key, default, default=default, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/.local/lib/python3.12/site-packages/pymemcache/client/hash.py", line 322, in _run_cmd
    return self._safely_run_func(client, func, default_val, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/.local/lib/python3.12/site-packages/pymemcache/client/hash.py", line 211, in _safely_run_func
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #645, opened 2025-05-30
FIX: optional sock

iwakitakuma33:
https://github.com/pinterest/pymemcache/issues/612


iwakitakuma33:
@jparise 
thanks for your review.

If the socket is disconnected at an inopportune moment, it must be handled appropriately on the Django side.
To that end, I made corrections to cause an error to occur.

PR #646, opened 2025-07-09
Added high-level diagrams

ivanmilevtues:
In this PR, I am adding a diagram visualizaiton for the git-stracktrace project.
I can see that you keep your `docs` as .rst files and build them with sphinx. I can integrate the diagrams within those if you like them (also we have a free github action to keep the whole thing up-to-date with new commits).

You can see how the diagrams in this change render here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/pymemcache/on_boarding.md

The goal of these documents is to help people get up to speed quickly with the git-stacktrace codebase. The diagrams are designed to give immediate overview of the codebase and allow you to dig deeper into each component of interest. The whole visualization is generated so it is easy to keep up-to-date. The generation is done via StaticAnalysis and LLMs.

Would love to connect and discuss how can we help the people who work and contributo to pinterest's codebase. We are curious to hear how do you keep docs up-to-date so that people can get up-to-speed to different projects within the company!

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Client_Wrappers_Extensions.md
+++ b/.codeboarding/Client_Wrappers_Extensions.md
@@ -1,0 +1,115 @@
+```mermaid
+
+graph LR
+
+    Client_API_Base_Client_["Client API (Base Client)"]
+
+    Hashing_Routing_Hash_Client_["Hashing/Routing (Hash Client)"]
+
+    Retry_Mechanism_Retrying_Client_["Retry Mechanism (Retrying Client)"]
+
+    AWS_ElastiCache_Extension["AWS ElastiCache Extension"]
+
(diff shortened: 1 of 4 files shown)


jparise:
Interesting idea, but we'll pass on this.

ivanmilevtues:
Hey @jparise, as you found the idea intersting I am dropping by to let you know that we are now open-source: https://github.com/CodeBoarding/CodeBoarding

If you want to check on it you can, i am happy to answer questions if any :-)

PR #647, opened 2025-08-20
MockMemcacheClient.set: assert that expire is an integer

snarfed:
the real client does this, so ideally the mock should too.

### DIFF
--- a/pymemcache/test/utils.py
+++ b/pymemcache/test/utils.py
@@ -102,6 +102,7 @@ class MockMemcacheClient:
 
         value, flags = self.serde.serialize(key, value)
 
+        assert isinstance(expire, int), expire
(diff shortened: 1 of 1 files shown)


jparise:
I think we'd ideally improve this via type checking rather than a runtime assertion that's only in the test code path. What do you think?

jparise (commented):
I think we'd ideally improve this via type checking rather than a runtime assertion that's only in the test code path. What do you think?

snarfed:
@jparise type annotation sounds great! I didn't go that route originally because MockMemcacheClient doesn't currently have any, and it felt awkward to add just one in isolation, and I wasn't quite ready to tackle the whole class. I'm open to anything though.

